PR #11, opened 2018-03-23
Added a check for an edge case with null values on logs

SilverXXX:
Found an edge case where a null value where passed to .Log, i added a check for it, returning and empty string field when found.
Found it while using opentracing-contrib/csharp-netcore

### DIFF
--- a/src/LetsTrace/FieldExtensions.cs
+++ b/src/LetsTrace/FieldExtensions.cs
@@ -51,33 +51,38 @@ namespace LetsTrace
             if (kv.Value is byte[]) {
                 return kv.Value.ToField<byte[]>(kv.Key);
             }
-
-            var typeCode = Type.GetTypeCode(kv.Value.GetType());
-            
-            switch(typeCode)
+            if (kv.Value == null) {
+                return new Field<string> { Key = kv.Key, Value = "" };
+            }
+            else
             {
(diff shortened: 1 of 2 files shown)


cwe1ss:
@SilverXXX Just curious, where did I add a null-value? I probably shouldn't do that either.

SilverXXX:
After a call to a webapi which actually had no arguments, in OpenTracingLogger

src/LetsTrace/FieldExtensions.cs
```diff
@@ -51,33 +51,38 @@ public static Field ToField(this KeyValuePair<string, object> kv)
...
                 return kv.Value.ToField<byte[]>(kv.Key);
             }
-
-            var typeCode = Type.GetTypeCode(kv.Value.GetType());
-            
-            switch(typeCode)
+            if (kv.Value == null) {
+                return new Field<string> { Key = kv.Key, Value = "" };
+            }
+            else
```
Falco20019: I would remove the else branch since the if-case already has a return in it. Reduces nesting by one level. And makes it similar to the byte-array case above.

Falco20019:
Thanks for your input. Please see my small remark.

grounded042:
Thanks @SilverXXX

PR #13, opened 2018-03-24
SpanBuilder must also look for an ActiveSpan in Start()

### LINKED ISSUE #12: SpanBuilder must also look for an ActiveSpan in Start()
Right now, only [`StartActive()`](https://github.com/Chatham/LetsTrace/blob/e680cf0a3679e5a40295e9ab386c84fb61ec3578/src/LetsTrace/SpanBuilder.cs#L55) takes an existing `tracer.ActiveSpan` as a parent. However, this must also be done in the regular `Start()` method. The only difference between `Start()` and `StartActive()` is that `StartActive()` *activates* the newly created span.

See: [Implementation in Jaeger](https://github.com/jaegertracing/jaeger-client-java/blob/d69c2d3333da79d7eb1ffdf982186de5b5f1f4bc/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java#L395) and [related upcoming specification](https://github.com/opentracing/specification/pull/108)

cwe1ss:
Fixes #12

### DIFF
--- a/src/LetsTrace/SpanBuilder.cs
+++ b/src/LetsTrace/SpanBuilder.cs
@@ -52,16 +52,21 @@ namespace LetsTrace
 
         public IScope StartActive(bool finishSpanOnDispose)
         {
-            if (_tracer.ScopeManager.Active != null && _references.Count == 0 && !_ignoreActiveSpan)
-            {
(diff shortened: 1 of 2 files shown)


grounded042:
Thanks @cwe1ss!

PR #16, opened 2018-03-24
AddReference() must ignore null contexts; FollowsFrom no longer exists

cwe1ss:
* `AddReference` / `AsChildOf` must ignore calls with null spans/contexts.
* I've also removed the `FollowsFrom` methods as they currently don't exist on the `ISpanBuilder` interface.

### DIFF
--- a/src/LetsTrace/SpanBuilder.cs
+++ b/src/LetsTrace/SpanBuilder.cs
@@ -32,7 +32,10 @@ namespace LetsTrace
 
         public ISpanBuilder AddReference(string referenceType, ISpanContext referencedContext)
         {
-            _references.Add(new Reference(referenceType, referencedContext));
+            if (referencedContext != null)
+            {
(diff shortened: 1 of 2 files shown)


src/LetsTrace/SpanBuilder.cs
```diff
@@ -32,7 +32,10 @@ public SpanBuilder(ILetsTraceTracer tracer, string operationName, ISampler sampl
 
         public ISpanBuilder AddReference(string referenceType, ISpanContext referencedContext)
         {
-            _references.Add(new Reference(referenceType, referencedContext));
+            if (referencedContext != null)
```
grounded042: Do we have a test case for this @cwe1ss?

cwe1ss:
I added some tests!

grounded042:
Thanks @cwe1ss!

PR #17, opened 2018-03-24
Shared Project settings; SourceLink; XML docs, TreatWarningsAsErrors

cwe1ss:
This adds a few infrastructure-related features:

* Directory.Build.props files for sharing common settings
* Enables TreatWarningsAsErrors and enabled latest C# version features (which we might need - e.g. readonly structs)
  * There was a warning with an override for Equals but not for GetHashCode. I'm not sure what the correct implementation is here as the Equals code is quite complex.
* Generates XML docs for all projects in \src
* Adds [SourceLink](https://github.com/ctaggart/SourceLink) support for all projects in \src 
  * This allows people to easily debug into the source code. This project is already part of the .NET Foundation and will probably be integrated directly into msbuild at some point.

The diff also contains some whitespace fixes... hope that's ok (VS Code automatically does that).

### DIFF
--- a/src/LetsTrace/Samplers/RemoteControlledSampler.cs
+++ b/src/LetsTrace/Samplers/RemoteControlledSampler.cs
@@ -24,21 +24,21 @@ namespace LetsTrace.Samplers
         private readonly Task _pollTimer;
 
         private RemoteControlledSampler(
-            string serviceName, 
+            string serviceName,
             ISamplingManager samplingManager,
-            ILoggerFactory loggerFactory, 
-            IMetrics metrics, 
+            ILoggerFactory loggerFactory,
+            IMetrics metrics,
             ISampler sampler,
             int pollingIntervalMs)
         : this(serviceName, samplingManager, loggerFactory, metrics, sampler, new SamplerFactory(), pollingIntervalMs, PollTimer)
         {}
 
         internal RemoteControlledSampler(
-            string serviceName, 
-            ISamplingManager samplingManager, 
-            ILoggerFactory loggerFactory, 
-            IMetrics metrics, 
-            ISampler sampler, 
+            string serviceName,
+            ISamplingManager samplingManager,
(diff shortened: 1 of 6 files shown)


src/LetsTrace/Samplers/RemoteControlledSampler.cs
```diff
@@ -192,6 +192,11 @@ public override bool Equals(object obj)
 
             return false;
         }
+        public override int GetHashCode()
+        {
+            // TODO Do we need special handling here? (see Equals handling)
+            return base.GetHashCode();
```
Falco20019: You can use the following code. When I started the method, I haven't had the readonly flags set and GetHashCode should only contain non-writeable data.

```csharp
unchecked
{
	var hashCode = (_samplingManager != null ? _samplingManager.GetHashCode() : 0);
	hashCode = (hashCode * 397) ^ (_serviceName != null ? _serviceName.GetHashCode() : 0);
	hashCode = (hashCode * 397) ^ (_loggerFactory != null ? _loggerFactory.GetHashCode() : 0);
	hashCode = (hashCode * 397) ^ (_logger != null ? _logger.GetHashCode() : 0);
	hashCode = (hashCode * 397) ^ (_metrics != null ? _metrics.GetHashCode() : 0);
	hashCode = (hashCode * 397) ^ (_samplerFactory != null ? _samplerFactory.GetHashCode() : 0);
	hashCode = (hashCode * 397) ^ (_cancellationTokenSource != null ? _cancellationTokenSource.GetHashCode() : 0);
	hashCode = (hashCode * 397) ^ (_pollTimer != null ? _pollTimer.GetHashCode() : 0);
	return hashCode;
}
```

grounded042:
Thanks @cwe1ss! We should get the hashcode functionality setup. If I have some time I'll take a look, but @Falco20019 feel free to submit another PR as it looks like you've already got something working.

PR #18, opened 2018-03-26
Fix DateTimeOffset tests

Falco20019:
DateTimeOffset.Parse was failing on non-english machines.


grounded042:
Thanks @Falco20019!

PR #19, opened 2018-03-26
Fixes xunit warning which is now an error and breaks the build

cwe1ss:
See latest [AppVeyor build output](https://ci.appveyor.com/project/chatham/letstrace/build/0.0.4-22-master#L118).

This probably is due to the introduction of TreatWarningsAsErrors=true.


grounded042:
Ah fun stuff. Thanks!

PR #22, opened 2018-03-26
Adds a build script

### LINKED ISSUE #20: Add a build script for local execution and use it in AppVeyor
Would be nice to have a `build.ps1` file which restores/builds/packs/tests locally. This would ensure we can submit PRs that don't fail on AppVeyor. AppVeyor should then also only call this script.

I'd be happy to send a PR with the script I'm using for several projects - e.g. in [opentracing-csharp](https://github.com/opentracing/opentracing-csharp/blob/master/build.ps1) if you like it. It's a simple script that doesn't have any dependencies on e.g. psake, cake, ... but still delivers a nice output. It's also faster than the current AppVeyor script as it does restore/build/pack in one call.

cwe1ss:
Resolves #20.

* There's now **only one place** where you have to update your version number after a release (Directory.Build.props). However, AppVeyor now no longer uses this number. I can change this back if you want.
* The script can be executed locally with `.\build.ps1` and you can use the parameters `-RunBuild $true` & `-RunTests $true` to e.g. not run tests etc.
* AppVeyor runs this script as well so there's only one place where the build pipeline is defined.
* AppVeyor sets a `VersionSuffix` if it is a regular commit/PR and it does NOT set it if it is a release (a git tag)

Just let me know if yo don't like something here!!

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,0 +1,77 @@
+[CmdletBinding(PositionalBinding = $false)]
+param(
+    [string] $VersionSuffix = "loc" + ((Get-Date).ToUniversalTime().ToString("yyyyMMddHHmm")),
+    [string] $ArtifactsPath = (Join-Path $PWD "artifacts"),
+    [string] $BuildConfiguration = "Release",
+
+    [bool] $RunBuild = $true,
+    [bool] $RunTests = $true
+)
+
+$ErrorActionPreference = "Stop"
+$Stopwatch = [System.Diagnostics.Stopwatch]::StartNew()
+
+function Task {
+    [CmdletBinding()] param (
+        [Parameter(Mandatory = $true)] [string] $name,
+        [Parameter(Mandatory = $false)] [bool] $runTask,
+        [Parameter(Mandatory = $false)] [scriptblock] $cmd
+    )
+
+    if ($cmd -eq $null) {
+        throw "Command is missing for task '$name'. Make sure the starting '{' is on the same line as the term 'Task'. E.g. 'Task `"$name`" `$Run$name {'"
+    }
+
+    if ($runTask -eq $true) {
+        Write-Host "`n------------------------- [$name] -------------------------`n" -ForegroundColor Cyan
+        $sw = [System.Diagnostics.Stopwatch]::StartNew()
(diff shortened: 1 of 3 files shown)


cwe1ss:
sorry - there's an issue with the appveyor script... I'll fix it...

cwe1ss:
I fixed the AppVeyor config.

build.ps1
```diff
@@ -0,0 +1,77 @@
...
+    Write-Host ".NET Core SDK: $(dotnet --version)`n"
+
+    Remove-Item -Path $ArtifactsPath -Recurse -Force -ErrorAction Ignore
+    New-Item $ArtifactsPath -ItemType Directory -ErrorAction Ignore | Out-Null
+    Write-Host "Created artifacts folder '$ArtifactsPath'"
+}
+
+Task "Build" $RunBuild {
+
+    dotnet msbuild "/t:Restore;Build;Pack" "/p:CI=true" `
```
grounded042: Not against using `dotnet msbuild`, but just wonder why you used that over using `dotnet build`?

cwe1ss: Because the dotnet CLI still is quite limited. It e.g. can't accept an empty VersionSuffix, so one would have to write a big `if ($VersionSuffix -eq $null) { dotnet build... } else { dotnet build --VersionSuffix $VersionSuffix }` with duplicate parameters. 

It also can't build & pack in one step which is a lot faster. With separate commands you again have to duplicate all the parameters etc.

Believe me, I've tried to exclusively use normal `dotnet` commands but I ran into so many issues (especially with older SDK versions) that this by far is the best way. This is also how it's done e.g. in [Dapper](https://github.com/StackExchange/Dapper/blob/07fe543f1530b33818c9ce989bf0672e145c954d/build.ps1#L123).

The regular `dotnet ***` commands [just call msbuild](https://github.com/dotnet/cli/blob/d0780d81e571abfcdcd1ed2ab85f0cc4a155528e/src/dotnet/commands/dotnet-build/BuildCommand.cs#L30) under the hood anyway.

grounded042: Gotcha. That makes sense.

PR #10, opened 2018-03-21
Merge LetsTrace.Jaeger

grounded042:

PR #26, opened 2018-03-26
Version bump; ignore artifacts folder

cwe1ss:
* 0.0.4 has already been released so all new builds should already have a higher version number.
* Added a general gitignore for the `artifacts/` folder so that nothing gets committed if someone e.g. extracts a nuget package for debugging cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -272,5 +272,9 @@ config.ps1
 # VS generated files
 launchSettings.json
 
+
+# Release Artifacts
+artifacts/
+
 # BenchmarkDotNet
 BenchmarkDotNet.Artifacts/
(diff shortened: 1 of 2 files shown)


grounded042:
We have a version bump sitting in #10 and once @kevindaviscfc is back in the office today to do a review of that PR we will be looking at getting that merged.

cwe1ss:
The version bump is no longer in that PR as so many things have changed due to my PRs (new AppVeyor script, Directory.Build.props etc).

grounded042:
Ah gotcha. Cool. I'd like to get that merged then and at that point we can push this through.

Issue #30, opened 2018-03-27
LetsTrace.Jaeger still contains `Field`


grounded042:
With the removal of `Field` in #24 the LetsTrace.Jaeger project broke. Lets fix that.

cwe1ss:
Will you fix it or should I? It should now be possible to use the `case byte[] value` syntax in the `JaegerFieldExtensions.Marshal` method.

grounded042:
I've got it open now and I'll take care of it. Was actually thinking of moving it to a static method since I don't feel like it's the best idea to put an extension method out there for `object`.

grounded042:
And I did remember you mentioned that new switch syntax which is so nice.

grounded042:
Unrelated to this specifically, but with moving ILRepack into the .csproj this no longer builds on MacOS for me. Not a big deal, but might be nice to figure that out. Not sure if there is a way to repack on MacOS.

cwe1ss:
What error do you get? I can have a look at this.

grounded042:
@cwe1ss the error is `LetsTrace/src/LetsTrace.Jaeger/LetsTrace.Jaeger.csproj(34,5): error MSB3073: The command "\.nuget\packages\ILRepack\2.0.15\tools\ilrepack.exe /parallel /internalize /out:LetsTrace.Jaeger.dll LetsTrace.Jaeger.dll Thrift.dll Jaeger.Thrift.dll" exited with code 127.`

It's probably because ILRepack is an exe that won't run on MacOS. Also for what it's worth Visual Studio on Windows doesn't understand that we are packing the dlls on build and so the LetsTrace.Jaeger.Tests project shows a bunch of errors for missing types that don't actually happen at compile time since DLLs are packed into the LetsTrace.Jaeger dll.

PR #31, opened 2018-03-27
remove `Field` support from LetsTrace.Jaeger

grounded042:
for #30


src/LetsTrace.Jaeger/Serialization/JaegerThriftSerialization.cs
```diff
@@ -84,10 +84,50 @@ public static List<JaegerTag> ConvertLogToJaegerTags(LogRecord log)
...
+                case double d:
+                    tags.Add(new JaegerTag{ Key = key, VType = JaegerTagType.DOUBLE, VDouble = d });
+                    break;
+                case decimal d:
+                    tags.Add(new JaegerTag{ Key = key, VType = JaegerTagType.DOUBLE, VDouble = (double)d });
+                    break;
+                case bool b:
+                    tags.Add(new JaegerTag{ Key = key, VType = JaegerTagType.BOOL, VBool = b });
+                    break;
+                case UInt16 u16:
```
cwe1ss: Nit: The case statements are now a mix between keywords (e.g. `string`) and types (e.g. `UInt32)`. You could also use `ushort`, `uint`, ... instead.

fyi: I'm not sure but I **think** you can also use the same variable for each case (e.g. `case string value:` and `case double value`). Not sure if this is more readable/better though.

test/LetsTrace.Jaeger.Tests/Serialization/JaegerThriftSerializationTests.cs
```diff
@@ -108,65 +107,61 @@ public void BuildJaegerLog()
...
-            Assert.Equal(boolField.Key, converted.Fields[2].Key);
-            Assert.Equal(boolField.Value, converted.Fields[2].VBool);
+            Assert.Equal(boolFieldKey, converted.Fields[2].Key);
+            Assert.Equal(fields[boolFieldKey], converted.Fields[2].VBool);
 
             Assert.Equal(JaegerTagType.LONG, converted.Fields[3].VType);
-            Assert.Equal(uint16Field.Key, converted.Fields[3].Key);
-            Assert.Equal(uint16Field.Value, converted.Fields[3].VLong);
+            Assert.Equal(uint16FieldKey, converted.Fields[3].Key);
+            Assert.Equal(Convert.ToInt64(fields[uint16FieldKey]), converted.Fields[3].VLong);
```
cwe1ss: Do you even have to use `Convert.ToInt64`? Doesn't the `Assert.Equal` statement also work without it?

grounded042: yeah, the `Assert.Equal` throws saying `Expected: 5 (System.UInt16) Actual:   5 (System.Int64)`. Unless you were referring to line 143 where I actually didn't need it.

src/LetsTrace.Jaeger/Serialization/JaegerThriftSerialization.cs
grounded042: Gotcha, yeah. It would be better to be more consistent and use the keywords.

test/LetsTrace.Jaeger.Tests/Serialization/JaegerThriftSerializationTests.cs
cwe1ss: ok, makes sense. thanks for the response!

src/LetsTrace.Jaeger/Serialization/JaegerThriftSerialization.cs
```diff
@@ -84,10 +84,50 @@ public static List<JaegerTag> ConvertLogToJaegerTags(LogRecord log)
...
+                    tags.Add(new JaegerTag{ Key = key, VType = JaegerTagType.BOOL, VBool = val });
+                    break;
+                case ushort val:
+                    tags.Add(new JaegerTag { Key = key, VType = JaegerTagType.LONG, VLong = val });
+                    break;
+                case uint val:
+                    tags.Add(new JaegerTag { Key = key, VType = JaegerTagType.LONG, VLong = val });
+                    break;
+                case ulong val:
+                    tags.Add(new JaegerTag{ Key = key, VType = JaegerTagType.LONG, VLong = (long)val });
```
cwe1ss: This cast will be a problem if the `ulong` is actually larger than the positive part of a `long`. I don't know if that throws an exception or just results in a different value.

Not sure if it's better to just not support `ulong` values?!

grounded042: It throws `Message: System.OverflowException : Value was either too large or too small for an Int64`

grounded042: I'd rather not get rid of it...what if we converted it to a string?

cwe1ss: yep, sounds good.

PR #33, opened 2018-03-27
add obsolete tag to let people know we are moving

grounded042:
had to not error on warning as the obsolete tag makes unit test builds to error - this is temp and should be changed once we make the move

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>0.0.5</VersionPrefix>
+    <VersionPrefix>0.0.6</VersionPrefix>
 
     <Authors>Chatham Financial Corp.</Authors>
     <Copyright>Copyright 2018 (c) Chatham Financial Corp. All rights reserved.</Copyright>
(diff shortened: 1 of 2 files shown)


cwe1ss:
Maybe its better to wait with this until we finished the migration and have a new release that people can move to immediately?! We can always create the "old" LetsTrace release by just making a branch from the last 0.0.4 release tag. This release will then have no changes from the last 0.0.4 release except the Obsolete tags (which contains links to the new URL and packages).

grounded042:
That makes sense. Do you know if AppVeyor would work with that if we just change up the config in that directory to tell AppVeyor to deploy a branch other than master?

cwe1ss:
As long as you build it with the existing AppVeyor account, the old API key should still work. I guess you still need owner permissions though for your AppVeyor account to still see the project once it's moved.

A simple alternative would be to manually deploy the packages from your local machine. This can be done very easily, now that we have `.\build.ps1`. We just have to prepare everything in a branch and then you can execute the following statement to create all nuget packages in the `.\artifacts\nuget` folder. You could then upload them manually on nuget.org
```ps
.\build.ps1 -VersionSuffix ""
```

However, If you want to merge & release this PR now, that should be fine as well as there's really not many users yet...

grounded042:
I'm going to look at pushing this up outside of CI/CD once we get the package up under the new name.

yurishkuro:
@grounded042 this PR looks like it's no longer relevant?

Falco20019:
This should only be build manually and pushed as update to the LetsTrace package to inform people about the new package.

cwe1ss:
Yep, I'm going to close this. Should @grounded042 want to do another release on the LetsTrace packages, he has to do that manually locally (or in a separate branch) anyway.

Issue #35, opened 2018-03-28
Thrift library has dependencies on Microsoft.AspNetCore.* and Microsoft.Extensions.* libraries


cwe1ss:
It seems like Thrift support for .NET is a huge mess. There's the original [csharp](https://github.com/apache/thrift/tree/master/lib/csharp) code base and there's also a newer [netcore](https://github.com/apache/thrift/tree/master/lib/netcore) code base. The latest `csharp` release is very old and doesn't target .NET Standard and the `netcore` project doesn't have any official NuGet releases yet.

This project currently has a local copy of the `netcore` project and it IL-merges it but even this project should not be used IMO because it has a a lot of weird non-standard dependencies - especially `Microsoft.AspNetCore`. This means that anyone who wants to use LetsTrace/Jaeger in a regular console app would get the whole ASP.NET Core stack DLLs as well - and this is not acceptable IMO.

We therefore need a different solution IMO. I can think of the following:

* As LetsTrace/Jaeger is currently targeting `netstandard2.0` we might be able to just reference the official `Thrift` package. They introduced this compatibility feature recently AFAIK. However, I'm not sure if the very old version of the official `Thrift` package still works for us.
* Vendor-in the `csharp` project instead of `netcore` and have it target `netstandard2.0`. Ideally, this should work with minimal effort as .NET Standard 2.0 is very compatible to the full .NET framework. We would still have to IL-merge it - or we could also just release it under a different name and save us the pain of IL-merging. (e.g. `Jaeger.Thrift.ThriftLib`).
* Try one of the existing forks of Thrift on NuGet - e.g. https://www.nuget.org/packages/Apache.Thrift/

Your thoughts?

cwe1ss:
I've been looking at another alternative: We can just remove the parts from the `netcore` code that depend on non-standard libraries. As far as I've seen that's just a handful of "Server" files. That will probably be the easiest solution.

grounded042:
Thrift has been a pain from the begging for all of the reasons you outlined above. I don't like IL-merging it, but it seemed like the best option at the time. It would be nice if they responded to the ticket that @Falco20019 put in (https://issues.apache.org/jira/browse/THRIFT-4512).

[...]

PR #36, opened 2018-03-28
Publish vendored Thrift as separate package without ASP.NET Core dependencies

cwe1ss:
See #35.

This PR does the following things:
* Thrift has been changed to no longer have dependencies on any non-standard libraries. I only had to remove two files and change two files for this (from the unused "Server" part and the project file itself) so merging changes from the official repo shouldn't be much harder.
* Thrift is no longer IL-merged as we are having issues with that on macOS and it also shows some errors in Visual Studio. Instead it is published as a separate package - currently named "Jaeger.Thrift.VendoredThrift" (I'm open to other naming suggestions). The assembly name will be the same, so there won't be any conflicts with other Thrift libraries.
* We should obviously remove that package as soon as there's a proper official Thrift library.

This is a bad solution as well so I'm not sure if we should do that. Instead of publishing the modified version as a separate package, I can also try to fix the IL-merge problems.


grounded042:
Wow, that was not a lot of code to be able to drop 9 dependencies.

Falco20019:
Have you checked if there is any other NuGET library for Thrift that targets `.netstandard2.0` that we could use? Would be the cleaner option. Didn't have a lot of time today, but if noone has checked, I would look it up tomorrow.

But interesting to see how little code is depending on those 9 packages. Apache should really pack those into a separate package as you proposed.

grounded042:
I did some poking around and am not finding a good library on NuGet for Thrift. I do like the way this is done - we are not making a Thrift package for anyone but are instead tying it to this project. I'm gonna chat with @kevindaviscfc about this PR, but I think we should do a merge and move forward.

grounded042:
@cwe1ss could you look at why the AppVeyor build is failing?

grounded042:
Once we get the PR build working we'll go ahead and merge this in. I chatted with @kevindaviscfc and we agree that this is not an ideal solution, but until Thrift publishes an official package and one that can break apart the server and client we should go with this approach.

PR #38, opened 2018-03-30
Jaeger rename

grounded042:
We will need to change links inside of the .csproj file as well as copyright info once the move has happened.


cwe1ss:
LGTM! 👍

PR #41, opened 2018-03-30
add DCO and CONTRIBUTING info

grounded042:


grounded042:
git config has the wrong email...hang on

PR #45, opened 2018-04-02
Update JaegerThriftTransport.cs

fangchd:
If BufferSize is 1, we should flush it immediately always intead of waiting for the next Append action.


grounded042:
Nice find @fangchd. Could you add a test for this? Also we'll need you to sign off your commits as outlined in [CONTRIBUTING.md](https://github.com/jaegertracing/jaeger-client-csharp/blob/master/CONTRIBUTING.md)

grounded042:
This looks good. Thanks @fangchd!

Issue #46, opened 2018-04-02
Add size checking in transport


grounded042:
Currently we do not check the buffer byte size in the `Jaeger.Transport.Thrift.Transport.JaegerThriftTransport` class. This can lead to the exception `Cannot flush closed transport. Message too long`. I have seen the exception when using the `JaegerUdpTransport`, but not the `JaegerHttpTransport` In other Jaeger client implementations there are checks to make sure exceptions like that do not happen: https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/com/uber/jaeger/senders/ThriftSender.java

We should add a check. Also of note is that `TMemoryBufferClientTransport` is missing a reset method which could complicate things as we need that.

yurishkuro:
not only client side exceptions - if the total UDP packet size exceeds 65000, the jaeger-agent will not be able to process it.

HTTP transport does not have these limitations.

grounded042:
Thanks for the info @yurishkuro. I'm hoping to tackle this today.

grounded042:
@yurishkuro is there a good way to determine the emit batch overhead? I ported `TestEmitBatchOverhead` to C# from https://github.com/jaegertracing/jaeger-client-go/blob/b7bacc465f385658dd3586f58810226776d2a4bf/transport_udp_test.go#L57 and I'm seeing a max overhead of 22. However, I'm unable to set the Client SeqId like in the go tests. I did not see anything in the Java client on how they determined the overhead. They have a different number `33` compared to the `30` in go, so I'm assuming it is not the same in each language.

yurishkuro:
Yes, the technique is different by language due to different thrift APIs. I think the simplest thing is to use BINARY encoding, instead of COMPACT encoding, to avoid variable size numbers which were the cause of the complexity in the Go implementation. I think the impact on the message size from compact encoding is really minimal.

grounded042:
Fixed in #47

PR #47, opened 2018-04-03
add size checking in UDP transport

grounded042:
This is for #46. We were not correctly checking UDP size and thus we could end up sending UDP packets that were too big. I broke out the buffering logic so that in the UDP transport we buffer items until we can't fit any more in a UDP packet. In the HTTP transport we buffer items until we reach the batch size.

### DIFF
--- a/src/Jaeger.Transport.Thrift/Transport/JaegerHTTPTransport.cs
+++ b/src/Jaeger.Transport.Thrift/Transport/JaegerHTTPTransport.cs
@@ -21,7 +21,7 @@ namespace Jaeger.Transport.Thrift.Transport
 
         /// <inheritdoc />
         /// <param name="batchSize">The number of spans to buffer before sending them to Jaeger</param>
-        /// <param name="host">The host the Jaeger is running on</param>
+        /// <param name="host">The host that Jaeger is running on</param>
         /// <param name="port">The port of the host that Jaeger will accept http thrift</param>
         public JaegerHttpTransport(int batchSize = 0, string host = TransportConstants.DefaultAgentHost, int port = TransportConstants.DefaultCollectorHttpJaegerThriftPort) 
             : this(new Uri($"http://{host}:{port}/api/traces"), batchSize)
--- a/src/Jaeger.Transport.Thrift/Transport/JaegerHTTPTransport.cs
+++ b/src/Jaeger.Transport.Thrift/Transport/JaegerHTTPTransport.cs
@@ -1,21 +1,73 @@
 using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Jaeger.Transport.Thrift.Serialization;
 using Jaeger.Transport.Thrift.Transport.Sender;
-using Thrift.Protocols;
+using JaegerSpan = Jaeger.Thrift.Span;
 
 namespace Jaeger.Transport.Thrift.Transport
 {
+    /// <inheritdoc />
+    /// <summary>
+    /// JaegerHttpTransport provides an implementation to transport spans over HTTP using
+    /// Thrift. Spans are buffered unitl reaching the BatchSize at which point all spans
+    /// in the buffer will be sent to Jaeger.
+    /// </summary>
     public class JaegerHttpTransport : JaegerThriftTransport
     {
         public Uri Uri { get; }
+        internal const int DefaultBatchSize = 100;
+        internal readonly int BatchSize;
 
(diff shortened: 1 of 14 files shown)


grounded042:
Actually, not quite ready to merge yet working through another bug

grounded042:
So, some fun stuff here...before I was working on Windows, but I was working through some example stuff on MacOS and I started getting the message to long error. Searching around it turns out that by default MacOS sets the maximum UDP package size to 9216 bytes. Running `sudo sysctl -w net.inet.udp.maxdgram=65535` fixes the problem until reboot and everything went along smoothly. So this is ready to merge - just be careful using UDP on MacOS.

Info from: https://stackoverflow.com/questions/22819214/udp-message-too-long

src/Jaeger.Transport.Thrift/Transport/JaegerHTTPTransport.cs
```diff
@@ -1,21 +1,73 @@
...
     {
         public Uri Uri { get; }
+        internal const int DefaultBatchSize = 100;
+        internal readonly int BatchSize;
 
-        public JaegerHttpTransport(string hostname = TransportConstants.DefaultAgentHost, int port = TransportConstants.DefaultCollectorHttpJaegerThriftPort, int bufferSize = 0) 
-            : this(new Uri($"http://{hostname}:{port}/api/traces"), bufferSize)
+        /// <inheritdoc />
+        /// <param name="batchSize">The number of spans to buffer before sending them to Jaeger</param>
+        /// <param name="host">The host the Jaeger is running on</param>
```
Falco20019: "The host __that__ Jaeger is running on"

src/Jaeger.Transport.Thrift/Transport/JaegerThriftTransport.cs
```diff
@@ -39,22 +35,43 @@ public void Dispose()
             _sender.Dispose();
         }
 
-        public async Task<int> AppendAsync(IJaegerCoreSpan span, CancellationToken canellationToken)
+        /// <summary>
+        /// AppendAsync serializes the passed in span into Jaeger Thrift and passes it off
+        /// to the sender which will buffer it to be sent. If the buffer is full enough it
+        /// will be flushed and the number of spans sent will be returned. If the buffer is
+        /// not full enough, nothing will be done a 0 will be returned to indicate that 0
```
Falco20019: "...nothing will be done __and__ a 0 will be returned to indicate that 0..."

src/Jaeger.Transport.Thrift/Transport/JaegerThriftTransport.cs
```diff
@@ -39,22 +35,43 @@ public void Dispose()
...
+        /// spans were sent.
+        /// </summary>
+        /// <param name="span">the span to serialize into Jaeger Thrift and buffer to be sent</param>
```
Falco20019: __T__he span to serialize ...

src/Jaeger.Transport.Thrift/Transport/JaegerThriftTransport.cs
```diff
@@ -39,22 +35,43 @@ public void Dispose()
...
+        /// <param name="cancellationToken"></param>
+        /// <returns>the number of spans flushed, if any</returns>
```
Falco20019: __T__he number of spans...

src/Jaeger.Transport.Thrift/Transport/JaegerUDPTransport.cs
```diff
@@ -1,24 +1,126 @@
...
+            if (_byteBufferSize <= _maxByteBufferSize) {
+                _sender.BufferItem(span);
+
+                // if we can potentially fit more spans in the buffer don't flush
+                if (_byteBufferSize < _maxByteBufferSize) {
+                    return 0;
+                }
+
+                // can't fit anything else in the buffer, flush it
+                return await FlushAsync(cancellationToken);
```
Falco20019: Looks like `_byteBufferSize` is not reset in this case. Maybe add `_byteBufferSize = _processByteSize`?

grounded042: Good catch. That woulda been a nice bug...

grounded042:
@Falco20019 thanks for the feedback. Fixed.

PR #53, opened 2018-04-11
Use separate files for TraceId and SpanId

cwe1ss:
They are not subclasses so they should be in separate files.


cwe1ss:
Thx @Falco20019 . 

I'll just go ahead and merge it - I guess there shouldn't be much need for discussion. Feel free to comment/raise an issue if you disagree.

Issue #29, opened 2018-03-27
Should this project have parity with jaeger-client-java?


cwe1ss:
If the plan is [to make this the official C# Jaeger client](https://github.com/jaegertracing/jaeger/issues/578), should this project be as close as possible to [`jaeger-client-java`](https://github.com/jaegertracing/jaeger-client-java)? I noticed that there's quite a few things missing or done differently right now. E.g. SpanContext is not immutable, metrics are implemented differently, there's `ILetsTrace*` interfaces here, etc.

This obviously makes porting features a little bit harder. For instance, I was trying to port the missing "baggage restrictions" files but this requires a few changes due to the non-immutability of SpanContext. Also, the tests can't be ported easily either.

I understand the desire to have your own implementation, so this would be fine by me. However, if you want to make it a Java-port, I could donate some time and bring this library on par with Java. I should have a few full days of availability until next week. This would obviously result in breaking changes though.

I could do the following things:
* Merge the LetsTrace.Jaeger PR
* Rename everything to `Jaeger.*` if you want (see  https://github.com/jaegertracing/jaeger/issues/578#issuecomment-376150297 )
* Remove the `ILetsTrace*` interfaces as they don't exist in Java
* Make SpanContext immutable
* Add missing features
* Bring existing features on par with jaeger-client-java where useful.

Let me know if this would help.

grounded042:
Thanks for the offer @cwe1ss. Your help is definitely appreciated. When we first started on this project we did not have any idea of becoming the official C# client for Jaeger - however, things change and it looks like we are moving that way (which I think is a good thing). 

We took a good bit of inspiration from the Java client in the beginning as well as the Go client, but we didn't do a straight port as being a Jaeger client was not the goal at the time. Internally we are working on adding LetsTrace to our stack, but the whole reason we open sourced it was because we thought the community could benefit from what we had and we knew we could benefit from what the community had to offer. I am still young in my C# experience and I'm sure there are better ways out there to approach different aspects in this library - all of that to say that we welcome community feedback and input on how to move this library forward and to better serve the community.

Personally I think that what you mentioned above is a good idea. I'd like to get some input from @Falco20019  and @yurishkuro on this before we move forward.

[...]

PR #54, opened 2018-04-11
Parity with Java for Metrics

cwe1ss:
I'm comparing the C# code with the Java code and I'm trying to get it as close as possible to make future changes easier. (#29)

This PR brings the "Metrics" feature closer to Java. However there still need to be a few differences:
* Java has a concrete class called `Metrics` which is used everywhere. Having a class with this name is not nice in C# because the namespace has the same name and people would have to use `Metrics.Metrics`. As I couldn't come up with a different name, I just kept the already existing `IMetrics` interface and moved Java's logic to a new `MetricsImpl` class.
* C# can't use complex types on attribute constructors. The current C# code already used some custom logic for this but I reverted this and simplified it. It's now pretty close to Java.

I also replaced the metrics unit tests with the current version from Java.


src/Jaeger.Core/Metrics/MetricsImpl.cs
```diff
@@ -0,0 +1,173 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+
+namespace Jaeger.Core.Metrics
+{
+    public class MetricsImpl : IMetrics
+    {
+        /// <inheritdoc/>
+        [Metric(name: "traces", tag1Key: "state", tag1Value: "started", tag2Key: "sampled", tag2Value: "y")]
```
yurishkuro: Is this a limitation of the annotation syntax that you can't do a dictionary?

If so I would recommend just having a string like `tags="state=started, sampled=y"` and parse it during construction. The objective is to make the annotation readable & easy, not to help the parsing engine (which isn't that complicated anyway).

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -28,7 +28,11 @@ private RemoteReporter(ITransport transport, ILoggerFactory loggerFactory, IMetr
             try
             {
                 int n = await _transport.CloseAsync(CancellationToken.None).ConfigureAwait(false);
-                _metrics.ReporterSuccess.Inc(n);
+
+                if (n > 0)
```
yurishkuro: does it matter if it's 0? seems like unnecessary complication of the code

test/Jaeger.Core.Tests/Reporters/RemoteReporterTests.cs
```diff
@@ -67,27 +67,15 @@ public void RemoteReporter_ShouldCallMetrics()
...
-        public void RemoteReporter_ShouldCallMetricsFactory()
-        {
-            var transport = Substitute.For<ITransport>();
-            var metricsFactory = InMemoryMetricsFactory.Instance;
-
-            using (var reporter = new RemoteReporter.Builder(transport).WithMetricsFactory(metricsFactory).Build())
-            {
-                Assert.IsType<InMemoryMetricsFactory.InMemoryElement>(reporter._metrics.ReporterSuccess);
-            }
+            metrics.ReporterSuccess.Received(1).Inc(2);
```
yurishkuro: why not use in-memory implementation and just assert the final counter value? With a mock you are testing not only external side effect, i.e. the resulting value of the counter which is what we're interested in, but also _how it gets there_, which is mostly an irrelevant implementation detail that makes the test itself more brittle.

src/Jaeger.Core/Metrics/MetricsImpl.cs
cwe1ss: Yes, this is a limitation in C#! I'll change it to your recommendation.

cwe1ss:
Superseded by #55.

PR #55, opened 2018-04-17
Parity with jaeger-client-java

cwe1ss:
As mentioned in #29, I ported almost everything from jaeger-client-java to C#. This results in some breaking changes and it also is a very big PR as I also reorderd methods etc. to be aligned with the Java code. This will make it much easier to port any future changes in jaeger-client-java to C# though.

The code here compiles and "works on my machine (TM)" but there's still a handful of failing tests. The problems are in the test code though as I haven't had the time to properly port & test some of the threading code in the unit tests. I'll obviously fix this before we merge this. I still already created this PR to get your feedback.

Highlights / important notes:
* `Tracer`, `Span`, `SpanBuilder`, `SpanContext` are now ports of Java.
    * E.g. SpanContext wasn't immutable. I'm also using ReadOnlyCollection/ReadOnlyDictionary in many places in Span/Tracer/SpanContext to ensure immutability (to reflect Java's `Collections.unmodifiableList`)
* Added "Baggage restrictions" feature (/baggage)
  * The code is very similar. We have to use different code for HTTP/JSON though. I introduced a `Util.IHttpClient` interface and a corresponding `Util.DefaultHttpClient` class for that purpose.
* Removed `IJaegerCore*` interfaces
  * The `Span`, `SpanBuilder` `Tracer` types are no longer used for mocking so we don't need these interfaces.
* "Metrics" code is now much closer to Java - see #54 for some more details. This PR also supersedes #54.
  * In C# a class shouldn't have the same name as a namespace as that makes it hard to use (`Metrics.Metrics`) so we have an `IMetrics` interface and a `MetricsImpl` class vs just a `Metrics` class in Java. 
  * C# doesn't allow complex types (e.g. maps) on attributes so I had to use string for the tags (as suggested by @yurishkuro in #54)
* "Propagation" code was quite different in C# - it's now a port of Java
  * C#'s generics are more restrictive than Java's though so one can't use e.g. `Injector<?>`. I solved this by using `object` in the general `IInjector`/`IExtractor`/`ICodec` interfaces and by providing strongly types `Injector`/`Extractor`/`Codec` base classes.
* "Reporters" have also been quite different in C# (there was a `ITransport` interface instead of `ISender` etc). I ported the Java code.
* "Samplers" have been similar already but I also made them a complete port.
* "Senders" were different in C# (they existed in the `Jaeger.Transport.Thrift` package. 
  * I merged that code with Java's "Senders" and integrated it into Jaeger.Core. 
  * This also means that the C# client now also uses a UdpSender by default.
* Java uses several lombok annotations. To reflect Java's \@value annotation, I introduced the `Util.ValueObject` base class.
* I ported the tests from Java and removed most tests from the existing C# code as it would have been too painful to update them.
* The examples now use `OpenTracing.Contrib.NetCore` for instrumentation. The previous C# code used the Jaeger-specific interfaces which is not a good idea as instrumentation code should always be vendor-independent. This keeps the example very small and lets it focus on how Jaeger must be set up.

Open issues/tasks/questions:
* C# still has separate `TraceId` and `SpanId` types. I quite like them as they allow us to better abstract the 64bit vs 128bit trace id issue etc. so I kept them for now. I changed the internals though as C# was using `ulong` (unsigned long) which didn't play well with the Thrift code etc. I have to do some more testing there.
* Once this PR is merged, I'd like to move everything from `Jaeger.Core` to `Jaeger` and remove the `Jaeger.Core` project. There's no longer a need for it and it just makes stuff more complex. I haven't done this in this PR yet as it would have completely messed up the diff view.

Again, I think it's early enough to do this. There's no benefit in C# having a different architecture where it's not necessary. Being as close to Java as possible allows us to innovate on both languages at the same time and with minimal additional effort.

My suggested process for this PR would be that you look at it in "reasonable" detail. Looking at every detail will probably be too painful. It might also be easier to just compare the code with Java. As soon as we've discussed/resolved any general/big issues, I'd suggest we just merge it, start using it and fix problems as we go...


yurishkuro:
Haven't reviewed yet but noticed your comment about thrift senders. In Java there's currently a bad dependency where core explicitly depends on the thrift module. The plan there is to use service loader pattern to allow Config to instantiate different components without pulling them all as dependencies.

Falco20019:
Thanks for the PR. I had a look at your branch on Monday. I think there is an easier way for generics with the injection interfaces since the type is used as one of the parameters. Also saw some other things I had some comments on :)

I‘m out of office today for some job interviews in Spain but have blocked some time tomorrow for the first feedback.

Falco20019:
I made some minor adjustments as commits. I tried some stuff with the `Injector` and `Extractor` out, but you are right, there is no good option. Either we move the `TCarrier` to the method and loose type-safety when supplying the `Codec` or in case of moving the `TCarrier` to the interface we would have problems getting the `PropagationRegistry` done in a similar way...

I would still keep the `Jaeger.Core` and `Jaeger.Transport.Thrift` packages separate. I think it's cleaner to not have a dependency in core on the senders. And as @yurishkuro mentions this is something that is also a bad thing where we are ahead of java.

The review needs some more time, I will continue it later today or tomorrow and submit it here later.

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -1,114 +1,222 @@
...
 
-        public async void Dispose()
+        public void Report(Span span)
         {
+            bool added = false;
             try
             {
-                int n = await _transport.CloseAsync(CancellationToken.None).ConfigureAwait(false);
-                _metrics.ReporterSuccess.Inc(n);
+                // Its better to drop spans, than to block here
```
Falco20019: It's better to drop spans than to block here (missing apostrophe and comma not needed)

cwe1ss:
> I made some minor adjustments as commits. 

Thanks for the fixes!

> I would still keep the Jaeger.Core and Jaeger.Transport.Thrift packages separate. I think it's cleaner to not have a dependency in core on the senders. 

That was fair when it was a tracer-independent library. It doesn't make much sense now IMO as Thrift is the only protocol used right now. Keeping this separate means that the library can't set up a default sender which is bad IMO as it's quite a fundamental deviation from Java.

> And as @yurishkuro mentions this is something that is also a bad thing where we are ahead of java.

But what if the Java solution will end up being "different"!? That's my main point with all of this: I fully understand how tempting it is to do stuff different but this only brings pain in the long term. A much better solution IMO would be to discuss this feature in the Java-repository with C# in mind and then just port the commit to C#. 
This ensures that we can focus on C#/.NET-specific stuff - e.g. like adding a library that makes it easier to use Jaeger with `Microsoft.Extensions.DependencyInjection` and `Microsoft.Extensions.Configuration` (to configure it from environment variables etc).

Falco20019:
I'm fine with keeping it close to the java version and discussing changes on a meta level. First set of review comments will come in in the next minues :)

src/Jaeger.Core/Samplers/RemoteControlledSampler.cs
```diff
@@ -159,99 +114,101 @@ internal void UpdatePerOperationSampler(PerOperationSamplingStrategies samplingP
...
+                    InitialSampler = new ProbabilisticSampler();
                 }
-                if (_metrics == null)
+                if (Metrics == null)
                 {
-                    _metrics = NoopMetricsFactory.Instance.CreateMetrics();
+                    Metrics = new MetricsImpl(NoopMetricsFactory.Instance);
                 }
-                return new RemoteControlledSampler(_serviceName, _samplingManager, _loggerFactory, _metrics, _initialSampler, _pollingIntervalMs);
+                return new RemoteControlledSampler(this);
```
Falco20019: 👍  I like that you just have a ctor for the `Builder`. Maybe we should use this everywhere where we only have a private ctor. Especially the `Tracer` ctor would become a lot easier and I'm sure some others as well.

src/Jaeger.Core/Senders/HttpSender.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        private readonly Agent.Client _agentClient;
+        private readonly THttpClientTransport _transport;
+
+        /// <summary>
+        /// </summary>
+        /// <param name="endpoint">Jaeger REST endpoint consuming jaeger.thrift, e.g
+        /// http://localhost:14268/api/traces</param>
+        /// <remarks>
+        /// Uses the default {@link okhttp3.OkHttpClient} which uses {@link okhttp3.ConnectionPool#ConnectionPool()}.
```
Falco20019: I'm quite sure that we don't use `okhttp3.OkHttpClient` ;)

src/Jaeger.Core/SpanContext.cs
```diff
@@ -1,54 +1,132 @@
...
 using OpenTracing;
 
 namespace Jaeger.Core
 {
-    // The SpanContext is used to pass the information needed by other spans so
-    // that they can correctly reference other spans
-    public class SpanContext : IJaegerCoreSpanContext
+    public class SpanContext : ISpanContext
     {
+        internal static readonly ReadOnlyDictionary<string, string> EmptyBaggage = new ReadOnlyDictionary<string, string>(new Dictionary<string, string>());
```
Falco20019: Use `IReadOnlyDictionary`. `Dictionary` already implements `IReadOnlyDictionary`. Wrapping in `ReadOnlyDictionary` not needed.

src/Jaeger.Core/SpanContext.cs
```diff
@@ -1,54 +1,132 @@
...
+            var newBaggage = new Dictionary<string, string>(Baggage);
+            if (value == null)
+            {
+                newBaggage.Remove(key);
+            }
+            else
+            {
+                newBaggage[key] = value;
+            }
+            return new SpanContext(TraceId, SpanId, ParentId, Flags, new ReadOnlyDictionary<string, string>(newBaggage), DebugId);
```
Falco20019: Use `IReadOnlyDictionary`. `Dictionary` already implements `IReadOnlyDictionary`. Wrapping in `ReadOnlyDictionary` not needed.

src/Jaeger.Core/SpanContext.cs
```diff
@@ -1,54 +1,132 @@
...
+        }
+
+        public SpanContext WithBaggage(Dictionary<string, string> newBaggage)
+        {
+            return new SpanContext(TraceId, SpanId, ParentId, Flags, new ReadOnlyDictionary<string, string>(newBaggage), DebugId);
```
Falco20019: Use `IReadOnlyDictionary`. `Dictionary` already implements `IReadOnlyDictionary`. Wrapping in `ReadOnlyDictionary` not needed.

src/Jaeger.Core/SpanContext.cs
```diff
@@ -1,54 +1,132 @@
...
+                newBaggage.Remove(key);
+            }
+            else
+            {
+                newBaggage[key] = value;
+            }
+            return new SpanContext(TraceId, SpanId, ParentId, Flags, new ReadOnlyDictionary<string, string>(newBaggage), DebugId);
+        }
+
+        public SpanContext WithBaggage(Dictionary<string, string> newBaggage)
```
Falco20019: Use `IReadOnlyDictionary`.

src/Jaeger.Core/SpanContext.cs
```diff
@@ -1,54 +1,132 @@
...
-            TraceId = traceId ?? throw new ArgumentNullException(nameof(traceId));
-            ParentId = parentId;
+        }
+
+        internal SpanContext(
+            TraceId traceId,
+            SpanId spanId,
+            SpanId parentId,
+            SpanContextFlags flags,
+            ReadOnlyDictionary<string, string> baggage,
```
Falco20019: Use `IReadOnlyDictionary`.

src/Jaeger.Core/Tracer.cs
```diff
@@ -18,189 +20,292 @@
...
+        public IClock Clock { get; }
         public IMetrics Metrics { get; }
+        public ILogger Logger { get; }
 
-        private Tracer(string serviceName, Dictionary<string, object> tags, IScopeManager scopeManager, ILoggerFactory loggerFactory,
-            IPropagationRegistry propagationRegistry, ISampler sampler, IReporter reporter, IMetrics metrics)
+        public IScopeManager ScopeManager { get; }
+        public ISpan ActiveSpan => ScopeManager.Active?.Span;
+
+        public ReadOnlyDictionary<string, object> Tags { get; }
```
Falco20019: Use `IReadOnlyDictionary`.

src/Jaeger.Core/Tracer.cs
```diff
@@ -18,189 +20,292 @@
...
+
+                    tags[Constants.TracerIpTagKey] = hostIPv4.ToString();
+                    IPv4 = Utils.IpToInt(hostIPv4);
+                }
+                catch
+                {
+                }
+            }
+
+            Tags = new ReadOnlyDictionary<string, object>(tags);
```
Falco20019: `Dictionary` already implements `IReadOnlyDictionary`. Wrapping in `ReadOnlyDictionary` not needed.

src/Jaeger.Core/Tracer.cs
```diff
@@ -18,189 +20,292 @@
...
-                    try
-                    {
-                        var hostIPv4 = System.Net.Dns.GetHostAddresses(hostname).First(ip => ip.AddressFamily == AddressFamily.InterNetwork).ToString();
-                        this.WithTag(Constants.TracerIpTagKey, hostIPv4);
-                    }
-                    catch
-                    {
-                    }
-                }
+                    TextMapCodec textMapCodec = new TextMapCodec(false);
```
Falco20019: Easier readable if ctor is inlined

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -1,114 +1,222 @@
...
+            }
+
+            public void Execute()
+            {
+                int n = _reporter._sender.Flush();
+                _reporter._metrics.ReporterSuccess.Inc(n);
             }
+        }
+
+        public sealed class Builder
```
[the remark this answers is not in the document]
yurishkuro: +1 - when you have a builder make the ctor private

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -1,114 +1,222 @@
...
         }
 
-        public class Builder
+        /*
+         * The code below implements the command pattern.  This pattern is useful for
+         * situations where multiple threads would need to synchronize on a resource,
+         * but are fine with executing sequentially.  The advantage is simplified code where
+         * tasks are put onto a blocking queue and processed sequentially by another thread.
+         */
+        public interface ICommand
```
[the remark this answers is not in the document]
cwe1ss: That one has a different signature. There's also no advantage in using it.

yurishkuro:
meta-comment: another issue I have with Java (and Go/Node) clients design is that I don't think the concept of the Sender is a good one. It would be better to have "transports" (udp, http, etc.) and "encoders" (thrift, json, proto), and in such a way that they could be mixed and matched.

src/Jaeger.Core/Reporters/RemoteReporter.cs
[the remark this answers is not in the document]
cwe1ss: My personal preference is to have stuff more separated as well, but I did it in the same way as Java because this way it's easier to compare files. The (potentially?!) better readability with multiple files was not worth deviating IMO.

I wonder how we should decide how to proceed here? Again, my vote is for being as close to Java as possible but I want to ensure we discuss this properly if you/anyone disagrees.

I'm writing this comment just once but it also applies to all other cases as well (e.g. Builder types).

src/Jaeger.Core/Samplers/ConstSampler.cs
```diff
@@ -1,31 +1,43 @@
...
+using System.Collections.ObjectModel;
+using Jaeger.Core.Util;
 
 namespace Jaeger.Core.Samplers
 {
-    // ConstSampler is a sampler that always makes the same decision.
-    public class ConstSampler : ISampler
+    public class ConstSampler : ValueObject, ISampler
     {
-        public bool Decision { get; }
```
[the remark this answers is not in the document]
cwe1ss: This property is currently not needed externally. There's also no public accessor in Java.

src/Jaeger.Core/Samplers/PerOperationSampler.cs
```diff
@@ -1,108 +1,145 @@
...
-                        _logger.LogInformation($"Exceeded the maximum number of operations ({_maxOperations}) for per operations sampling");
+                        if (OperationNameToSampler.Count < _maxOperations)
+                        {
+                            sampler = new GuaranteedThroughputSampler(samplingRate, LowerBound);
+                            OperationNameToSampler[operation] = sampler;
+                            isUpdated = true;
+                        }
+                        else
+                        {
+                            _logger.LogInformation("Exceeded the maximum number of operations {maxOperations} for per operations sampling", _maxOperations);
```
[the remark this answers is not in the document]
cwe1ss: No - see https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.1&tabs=aspnetcore2x

src/Jaeger.Core/Samplers/PerOperationSampler.cs
```diff
@@ -1,108 +1,145 @@
...
+                    sampler.Dispose();
+                }
+            }
+        }
+
+        protected override IEnumerable<object> GetAtomicValues()
+        {
+            yield return DefaultSampler;
+            yield return LowerBound;
+            yield return _maxOperations;
```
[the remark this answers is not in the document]
cwe1ss: +1. I'll update it as Java has `@Getter(AccessLevel.PACKAGE)` on the type and therefore also has an internal accessor.

src/Jaeger.Core/Samplers/ProbabilisticSampler.cs
```diff
@@ -1,45 +1,70 @@
...
+                _negativeSamplingBoundary = (long)((1L << 63) * samplingRate);
+            }
+
+            _tags = new ReadOnlyDictionary<string, object>(new Dictionary<string, object> {
+                { Constants.SamplerTypeTagKey, Type },
+                { Constants.SamplerParamTagKey, samplingRate }
+            });
+        }
+
+        public virtual SamplingStatus Sample(string operation, TraceId id)
```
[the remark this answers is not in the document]
cwe1ss: The code is the same as in Java. Wasn't worth deviating IMO.

Testing this logic is still on my todo list. :)

src/Jaeger.Core/Reporters/RemoteReporter.cs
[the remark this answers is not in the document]
cwe1ss: The code is the same as in Java - the ctor is package-internal as it's used in some unit tests. Sure, making it private and using the builder everywhere would be better but that wasn't my goal of this PR. We can of course improve this in Java and here in a separate PR.

src/Jaeger.Core/Samplers/RateLimitingSampler.cs
```diff
@@ -1,51 +1,54 @@
...
-    // RateLimitingSampler creates a sampler that samples at most maxTracesPerSecond. The distribution of sampled
-    // traces follows burstiness of the service, i.e. a service with uniformly distributed requests will have those
-    // requests sampled uniformly as well, but if requests are bursty, especially sub-second, then a number of
-    // sequential requests can be sampled each second.
-    public class RateLimitingSampler : IRateLimitingSampler
-    {
-        internal readonly IRateLimiter _rateLimiter;
-        private readonly Dictionary<string, object> _tags;
+        private readonly RateLimiter _rateLimiter;
+        private readonly ReadOnlyDictionary<string, object> _tags;
```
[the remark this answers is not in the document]
cwe1ss: That's a private variable - there's no reason to use an interface here.

src/Jaeger.Core/Samplers/SamplingStatus.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Jaeger.Core.Util;
+
+namespace Jaeger.Core.Samplers
+{
+    public sealed class SamplingStatus : ValueObject
+    {
+        public bool IsSampled { get; }
+        public ReadOnlyDictionary<string, object> Tags { get; }
```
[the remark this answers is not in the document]
cwe1ss: +1; I'll change this as Samplers are an extension point so the API should use the interface.

src/Jaeger.Core/Span.cs
```diff
@@ -1,126 +1,259 @@
...
+        }
 
-            OperationName = operationName;
-            Context = context ?? throw new ArgumentNullException(nameof(context));
-            StartTimestampUtc = startTimestampUtc ?? Tracer.Clock.UtcNow();
-            Tags = tags ?? new Dictionary<string, object>();
-            References = references ?? Enumerable.Empty<Reference>();
+        public ReadOnlyCollection<Reference> GetReferences() => _references;
+
+        public ReadOnlyDictionary<string, object> GetTags() => new ReadOnlyDictionary<string, object>(_tags);
```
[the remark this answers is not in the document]
cwe1ss: I'm not sure we should expose the interfaces on Span & SpanContext as that's the critical path. Using the interface has a (very) small performance cost and there's no real advantage here so I'd rather keep the concrete types.

src/Jaeger.Core/Propagation/TextMapCodec.cs
```diff
@@ -0,0 +1,174 @@
...
+        private TextMapCodec(Builder builder)
+        {
+            _urlEncoding = builder.UrlEncoding;
+            _contextKey = builder.SpanContextKey;
+            _baggagePrefix = builder.BaggagePrefix;
+        }
+
+        public override string ToString()
+        {
+            StringBuilder buffer = new StringBuilder();
```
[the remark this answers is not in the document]
cwe1ss: I ported it from Java. I have to use this special logic for `_urlEncoding` as a bool.ToString() returns `True` in C# vs `true` in Java. It would get less readable due to that. Would you still prefer the interpolation syntax?

src/Jaeger.Core/Propagation/TextMapCodec.cs
```diff
@@ -0,0 +1,174 @@
...
+        protected override SpanContext Extract(ITextMap carrier)
+        {
+            SpanContext context = null;
+            Dictionary<string, string> baggage = null;
+            string debugId = null;
+
+            foreach (var entry in carrier)
+            {
+                // TODO there should be no lower-case here
+                string key = entry.Key.ToLowerInvariant();
```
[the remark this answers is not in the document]
cwe1ss: That's a TODO from Java... We should look at it in a separate issue.

src/Jaeger.Core/Reporters/LoggingReporter.cs
```diff
@@ -4,17 +4,25 @@
...
+
         public LoggingReporter(ILoggerFactory loggerFactory)
         {
             Logger = loggerFactory?.CreateLogger<LoggingReporter>() ?? throw new ArgumentNullException(nameof(loggerFactory));
         }
 
-        public void Dispose() {}
+        public void Report(Span span)
+        {
+            Logger.LogInformation("Span reported: {span}", JsonConvert.SerializeObject(span));
```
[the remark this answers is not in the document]
cwe1ss: the usage is correct. see previous comment.

src/Jaeger.Core/Samplers/RemoteControlledSampler.cs
cwe1ss: I agree - I'd prefer to do this in a separate PR though as this should also be done in Java.

cwe1ss:
> @yurishkuro meta-comment: another issue I have with Java (and Go/Node) clients design is that I don't think the concept of the Sender is a good one. It would be better to have "transports" (udp, http, etc.) and "encoders" (thrift, json, proto), and in such a way that they could be mixed and matched.

I agree. My goal of this PR was to make the C# implementation exactly as good/bad as the Java implementation as fast as possible and to be able to synchronize future changes. We should discuss any improvements (which should be done in Java and here) in separate issues IMO.

src/Jaeger.Core/Samplers/RateLimitingSampler.cs
```diff
@@ -1,51 +1,54 @@
...
         public RateLimitingSampler(double maxTracesPerSecond)
-            : this(maxTracesPerSecond, new RateLimiter(maxTracesPerSecond, Math.Max(maxTracesPerSecond, 1.0)))
-        {}
-
-        public RateLimitingSampler(double maxTracesPerSecond, IRateLimiter rateLimiter)
         {
             MaxTracesPerSecond = maxTracesPerSecond;
+            _rateLimiter = new RateLimiter(maxTracesPerSecond, Math.Max(maxTracesPerSecond, 1.0));
+
+            _tags = new ReadOnlyDictionary<string, object>(new Dictionary<string, object> {
```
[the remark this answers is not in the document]
cwe1ss: Hmm. Having the strong guarantee with the extra `ReadOnlyDictionary` object might not be necessary though... If we just use the regular `Dictionary`, as you've suggested, someone would have to cast the `IReadOnlyDictionary` back to a `Dictionary` to manipulate it. That's not really a scenario we need to protect us from in this library so we might as well save some memory and return the objects themselves.

cwe1ss:
@Falco20019 I addressed your feedback. As I've written in https://github.com/jaegertracing/jaeger-client-csharp/pull/55#discussion_r182846385, the code now only uses the `IReadOnly*` interfaces to ensure immutability. I guess that also gives us better performance than allocating a new object and using the concrete `ReadOnly*` types.

I hope to have some time this weekend to fix the remaining unit tests and to do more testing.

src/Jaeger.Core/Samplers/RateLimitingSampler.cs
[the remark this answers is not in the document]
Falco20019: I assume you already saw that the `IReadOnlyDictionary` interface does not have an `Add` method and that the throwing code in the `ReadOnlyDictionary` class is for fulfilling the `IDictionary` interface. So it's all just about saving memory and runtime :) I saw that you already commited it, I just wanted to state it here again in case someone reads it up.

src/Jaeger.Core/Reporters/RemoteReporter.cs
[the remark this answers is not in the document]
Falco20019: Haven't thought about the tests :) But I still find it somewhat ugly that it's mixed. So maybe also something to have as issue on the meta level.

Falco20019:
Looks a lot better now! The `IReadOnly*` usage made a lot of places like [this one](https://github.com/jaegertracing/jaeger-client-csharp/pull/55/commits/b16c60631145ffe81d5ce39413b54d8e445f5cf6#diff-f4b703f3fa1ebd7a26e93e24b2d3c00eR102) a lot easier to read. I continue the review now and will have a look at all the baggage and other new code and go through the codebase with ReSharper to get some more feedback for optimization.

src/Jaeger.Core/Util/DateTimeExtensions.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Jaeger.Core.Util
 {
-    public static class DateTimeExtensionMethods
+    public static class DateTimeExtensions
```
Falco20019: I would prefer to use `DateTimeOffset` in the whole code base instead of mixing `DateTimeOffset` with `DateTime`. The code base looks like it uses Utc internally correct but still, it's unnecessary and bears a lot of bug potential (and needs this unnecessary extension class).

If you are fine with it, I could also submit the code for it. Especially, the `ToUnixTimeMilliseconds` method of `DateTimeOffset` uses UTC, independant of whatever timezone the user created the time for.

Falco20019: Ah, seems like I missed that issue completely. Thanks also for providing the benchmark over there! Pretty interesting that it's that much slower compared to simple `DateTime` instances.

cwe1ss:
I went through another iteration and (hopefully) fixed all but one flaky RemoteReporter tests. I think the problem there is in the test code and not in the library and I already invested a few hours, so I'm just skipping this test for now. Maybe someone else can look into this at some point.

I also added the CancellationToken-logic from the existing `ITransport` back to `ISender` to have a proper Close/Dispose story. This has to be different in C# due to async/await.

I also added `ConfigureAwait(false)` on all `await`s because I thought it's related to the flaky tests and it's recommended for libraries anyway (that are used outside of ASP.NET Core, which wouldn't need it).

I think I'm done now. @Falco20019 looking forward to your approval or further feedback.

cwe1ss:
I now also ported the "Configuration" class which should have been the last missing feature from Java.

src/Jaeger.Core/Propagation/TextMapCodec.cs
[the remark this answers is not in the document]
Falco20019: I also find it ugly that bools are upper-case in C# (especially since the idea is that it's XML compliant ant XML specifies it lower-case...). I assume `$"TextMapCodec{{contextKey={_contextKey},baggagePrefix={_baggagePrefix},urlEncoding={_urlEncoding.ToString().ToLowerInvariant()}}}"` to still be faster than to use a `StringBuilder` with a lot of Appends, but I haven't tested it.

Maybe we could extract the `var urlEncodedLower = _urlEncoding.ToString().ToLowerInvariant()` to keep the interpolated string readable. I at least find interpolation syntax easier readable than `StringBuilder` syntax.

src/Jaeger.Core/Senders/ISender.cs
```diff
@@ -1,11 +1,15 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
 namespace Jaeger.Core.Senders
 {
     public interface ISender
     {
-        int Append(Span span);
+        Task<int> AppendAsync(Span span, CancellationToken cancellationToken);
```
Falco20019: We should use `CancellationToken cancellationToken = default` as parameter. According to MSDN, `default(CancellationToken)` is semantically equivalent to `CancellationToken.None` and the source code shows, that  `CancellationToken.None` just does exactly return `default(CancellationToken)`. With this, the implementors of the interface don't have to offer both variants and `AppendCommand` and `FlushCommand` in `RemoteReporter` don't have to use `CancellationToken.None` directly, which always looks like a workaround.

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -62,52 +60,31 @@ public void Report(Span span)
...
-                _commandQueue.CompleteAdding();
+                // Give processor some time to process any queued commands.
 
-                // Give sender some time to process any queued spans.
-                _queueProcessor.Wait(10000);
+                var cts = CancellationTokenSource.CreateLinkedTokenSource(
+                    cancellationToken,
+                    new CancellationTokenSource(10000).Token);
+
+                _queueProcessorTask.Wait(cts.Token);
```
Falco20019: `_flushTask` is never awaited. Maybe we should await it here too after the `_queueProcessorTask`? The `_commandQueue.CompleteAdding()` above should lead the `FlushLoop` to be closed after max. `_flushInterval`. Also I don't like to have a sync `Wait` in an async method. I would propose the following:

```csharp
var cts = CancellationTokenSource.CreateLinkedTokenSource(
    cancellationToken,
    new CancellationTokenSource(TimeSpan.FromSeconds(10)).Token);
var cancellationTask = Task.Delay(Timeout.Infinite, cts.Token);

await Task.WhenAny(_queueProcessorTask, cancellationTask);
await Task.WhenAny(_flushTask, cancellationTask);
```

The `Task.Delay` variant to create a cancellationTask can also be found in the [MSDN Blog](https://blogs.msdn.microsoft.com/cesardelatorre/2017/11/18/implementing-background-tasks-in-microservices-with-ihostedservice-and-the-backgroundservice-class-net-core-2-x/). Another option would be the [CancellationTokenTaskSource](https://github.com/StephenCleary/AsyncEx/blob/master/src/Nito.AsyncEx.Tasks/CancellationTokenTaskSource.cs) from Stephen Cleary. A third option would be to not use `CreateLinkedTokenSource` but only `Task.Delay(10000, cancellationToken)` and check if the result of `await Task.WhenAny` is `cancellationTask`, but in that case no `OperationCanceledException` would be thrown on timeout, only on `cancellationToken`. That's why I prefer the first two options.

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -160,9 +165,9 @@ public AppendCommand(RemoteReporter reporter, Span span)
                 _span = span;
             }
 
-            public void Execute()
+            public Task ExecuteAsync()
             {
-                _reporter._sender.Append(_span);
+                return _reporter._sender.AppendAsync(_span, CancellationToken.None);
```
Falco20019: I don't like to have `CancellationToken.None` in library code since it somewhat breaks the token flow. Usually it's better to just use the method variant without `CancellationToken`. Since this method is defined in `ISender` and without alternative, I proposed to change `ISender` to use `CancellationToken cancellationToken = default` above. The implementing classes don't have to be changed than and it behaves exactly as with `CancellationToken.None`  being used here.

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -175,9 +180,9 @@ public FlushCommand(RemoteReporter reporter)
                 _reporter = reporter;
             }
 
-            public void Execute()
+            public async Task ExecuteAsync()
             {
-                int n = _reporter._sender.Flush();
+                int n = await _reporter._sender.FlushAsync(CancellationToken.None).ConfigureAwait(false);
```
Falco20019: See other comment on `CancellationToken.None` above.

src/Jaeger.Core/Tracer.cs
```diff
@@ -165,7 +181,7 @@ public sealed class Builder
             private ISampler _sampler;
             private IReporter _reporter;
             private PropagationRegistry _registry;
-            private IMetrics _metrics;
+            private IMetrics _metrics = new MetricsImpl(NoopMetricsFactory.Instance);
```
Falco20019: Don't define it here, since `Build` has the fallback code which we need anyway in case someone calls `WithMetrics(null)`.

src/Jaeger.Core/Tracer.cs
```diff
@@ -288,7 +299,7 @@ public Builder WithTag(string key, string value)
                 return this;
             }
 
-            private Builder WithTags(IDictionary<string, object> tags)
+            public Builder WithTags(IEnumerable<KeyValuePair<string, string>> tags)
```
Falco20019: Should we keep a variante with `IEnumerable<KeyValuePair<string, object>>` or are we sure we will only have string pairs? I assume this was changed for `Configuration.GetTracerBuilder`.

test/Jaeger.Core.Tests/Reporters/RemoteReporterTests.cs
```diff
@@ -232,33 +243,51 @@ public void TestFlushUpdatesQueueLength()
...
+            ((RemoteReporter)_reporter).Flush();
 
             Assert.True(_metricsFactory.GetGauge("jaeger:reporter_queue_length", "") > 0);
         }
 
-        [Fact(Skip = "Doesn't yet work properly")]
+        [Fact]
         public void TestFlushIsCalledOnSender()
         {
+            CountdownEvent countdownEvent = new CountdownEvent(1);
```
Falco20019: When running alone, this seems to succeed. When I run the whole session, this most of the time fails. Looks like there are some side effects on other tests.

src/Jaeger.Core/Configuration.cs
```diff
@@ -0,0 +1,783 @@
...
+            }
+
+            // for tests
+            internal ISampler CreateSampler(string serviceName, IMetrics metrics)
+            {
+                string samplerType = StringOrDefault(Type, RemoteControlledSampler.Type);
+                double samplerParam = Param.GetValueOrDefault(ProbabilisticSampler.DefaultSamplingProbability);
+                string hostPort = StringOrDefault(ManagerHostPort, HttpSamplingManager.DefaultHostPort);
+
+                if (string.Equals(samplerType, ConstSampler.Type, StringComparison.Ordinal))
```
Falco20019: Maybe use a switch on `samplerType`

src/Jaeger.Core/Reporters/CompositeReporter.cs
```diff
@@ -13,17 +15,17 @@ public CompositeReporter(params IReporter[] reporters)
...
             {
                 reporter.Report(span);
             }
         }
 
-        public void Dispose()
+        public async Task CloseAsync(CancellationToken cancellationToken)
         {
-            foreach(var reporter in _reporters)
+            foreach (var reporter in _reporters)
```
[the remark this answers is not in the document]
cwe1ss: Hmm.. I'm not really happy with the readability of this code. As this is in `CloseAsync` (which is only called once on app shutdown) the potential minimal performance gain of saving one state machine might not be worth the more complex code?!

If it were on `Report` (and if it were proven to be faster), I'd be ok with it of course.

src/Jaeger.Core/Tracer.cs
cwe1ss: I changed this because Java also only has a `String, String` method.

Falco20019: Still looks strange since the receiving dictionary is <string, object>. Is java using <string, string>? (Can’t look right now, just boarding my plane to Graz).

src/Jaeger.Core/Senders/ISender.cs
cwe1ss: I disagree on this one. While optional parameters are quite nice on concrete classes, they are confusing as hell on interfaces - see e.g. [link 1](https://blogs.msdn.microsoft.com/ericlippert/2011/05/09/optional-argument-corner-cases-part-one/), [link 2](http://blackrabbitcoder.net/archive/2010/06/17/c-optional-parameters---pros-and-pitfalls.aspx), [link 3](https://stackoverflow.com/questions/4922714/why-are-c-sharp-4-optional-parameters-defined-on-interface-not-enforced-on-imple#4923642). There's even a code analysis rule that warns when optional parameters are used on public types: [CA1026](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1026-default-parameters-should-not-be-used).

The thing with making the cancellation token optional is that as soon as that happens, no one uses it anymore. Keeping it as required forces people to either supply it or to make the conscious  decision to ignore it (by passing `CancellationToken.None`). This also makes it much easier to search for such cases because you can always search for "CancellationToken.None".

We should rather look at my current usages of `CancellationToken.None` and check if we should pass some timeout instead.

src/Jaeger.Core/Reporters/RemoteReporter.cs
cwe1ss: See my previous comment on why I disagree. Instead we should decide if it makes sense to provide an actual timeout-token here.

src/Jaeger.Core/Tracer.cs
cwe1ss: I don't like this either. This was done for parity with Java. I want to create a separate PR that fixes this behavior for all builders afterwards (for Java and C#).

src/Jaeger.Core/Tracer.cs
cwe1ss: yes, they are using <string, string>.

test/Jaeger.Core.Tests/Reporters/RemoteReporterTests.cs
cwe1ss: I hate multi threading tests! 😞

src/Jaeger.Core/Configuration.cs
cwe1ss: I will change it.

cwe1ss:
@Falco20019 I addressed your feedback. Regarding CancellationToken.None: I think we should discuss in a separate issue if/what timeout we choose for those situations (it's only used in the commands which are on a background thread). Right now it's basically the same behavior there as in Java - we block.

@Falco20019 @yurishkuro How should we handle the squashing of this PR? There are a few commits from @Falco20019 so if I manually do a squash or if use GitHubs "Squash and merge", those commits are probably no longer attributed to him. @Falco20019 would you be ok with that?

src/Jaeger.Core/Tracer.cs
Falco20019: It‘s ok to keep it for the PR but this is also unnecessary code in Java since it‘s redundant and defies the purpose of the setter since it can’t become null if not set explicitly.

Falco20019:
I‘m fine with squashing. I think nearly everything was just ReSharper stuff anyway, nothing to be contributed for :) Everything else I saw was discussed as comments and not changed directly. The rest of my comments on Configuration should come tomorrow.

src/Jaeger.Core/Configuration.cs
```diff
@@ -0,0 +1,775 @@
...
+            int? valueInMs = GetPropertyAsInt(name, logger);
+            if (valueInMs.HasValue)
+            {
+                return TimeSpan.FromMilliseconds(valueInMs.Value);
+            }
+            return null;
+        }
+
+        /// <summary>
+        /// Gets the system property defined by the name , and returns a boolean value represented by
```
Falco20019: ...defined by the name[- ], and returns a... (whitespace too much)

src/Jaeger.Core/Configuration.cs
```diff
@@ -0,0 +1,775 @@
...
+        /// the name. This method defaults to returning false for a name that doesn't exist.
+        /// @param name The name of the system property
```
Falco20019: This needs to be <param> not Java notation

src/Jaeger.Core/Configuration.cs
```diff
@@ -0,0 +1,775 @@
...
+        /// </summary>
+        private static bool? GetPropertyAsBool(string name, ILogger logger)
+        {
+            string value = GetProperty(name);
+            if (!string.IsNullOrEmpty(value))
+            {
+                if (string.Equals(value, "1", StringComparison.Ordinal))
```
Falco20019: Maybe use a switch-statement

Falco20019:
@cwe1ss If you do not have anything left you want to change, I would propose to merge it soon. I'm waiting for the next version since this fixes some of my problems right now like adding complex objects to logs which are now strapped. The rest of my feedback were just minor things, nothing functional.

cwe1ss:
@Falco20019 I've addressed your feedback. As there hasn't been anyone else commenting here, I'll just go ahead and merge it to move this forward. We should do another review of the code base then. I'll als try to create some separate PRs with stuff I'd like to discuss/change.

Issue #44, opened 2018-04-01
Extend OpenTracing tutorials with .NET + Jaeger example


yurishkuro:
https://github.com/yurishkuro/opentracing-tutorial

Falco20019:
If anyone want's to tackle it, feel free! I will have a look at it on friday if the issue is still unresolved :)

Falco20019:
Takes some more time then expected, since the lessons 3-4 are not so trivial to keep simple and understandable without too much ASP.NET Core code. Will have to continue next week.

Falco20019:
I internally assigned the issue to my colleague @saschaholesch since it's a good opportunity to get him into OpenTracing and prepare the tutorial also for other of my colleagues.

Falco20019:
This will be solved by merging https://github.com/saschaholesch/opentracing-tutorial/ (just awaiting #62).

Falco20019:
This will be fixed by https://github.com/yurishkuro/opentracing-tutorial/pull/21

PR #59, opened 2018-05-14
Move everything from Jaeger.Core into Jaeger

cwe1ss:
As announced briefly in #55, I'd like to simplify the package structure by moving everything from `Jaeger.Core` into `Jaeger` and by removing `Jaeger.Core` as a result. There's really no need right now to have this separation IMO.

This would simplify the solution and reduce confusion (e.g. people currently reference the `Jaeger` package but all types are in the `Jaeger.Core` namespace)

The corresponding jaeger-client-java package is called "jaeger-core" but the java package name is "io.jaegertracing" so that one doesn't include "core" either.


Falco20019:
The readme is still having the old names. But since there is #44, it should be fine.

cwe1ss:
@Falco20019 Thx for the review, I fixed the wrong namespaces!

I'll go ahead and merge this now!

PR #60, opened 2018-05-16
Set trace id via SpanBuilder

Marfusios:
Hello,

we have a system built on Kubernetes and [Linkerd ](https://linkerd.io/) and we need to somehow merge tracing logs inside the Jaeger from an infrastructure layer and from an application layer. 

Linkerd already sends trace id in an HTTP header, so it would be great to have the possibility to override trace id while building Span on the application layer. 
Is the provided solution convenient?

### DIFF
--- a/src/Jaeger.Core/SpanBuilder.cs
+++ b/src/Jaeger.Core/SpanBuilder.cs
@@ -14,6 +14,7 @@ namespace Jaeger.Core
         private DateTime? _startTimestampUtc;
         private List<Reference> _references;
         private bool _ignoreActiveSpan;
+        private TraceId? _traceId;
 
         internal SpanBuilder(Tracer tracer, string operationName)
         {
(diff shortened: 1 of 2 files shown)


Falco20019:
If you can bring Linkerd to use the format that OpenTracing uses for SpanContext, then you can use the normal extraction process. Right now, you can create you own SpanContext with a specific tracing id using [SpanContext.ContextFromString(string)](https://github.com/jaegertracing/jaeger-client-csharp/blob/master/src/Jaeger.Core/SpanContext.cs#L65) and let the new trace from the `SpanBuilder` reference that one. This would allow you to stay in the regular OpenTracing portability. This would still lock you on to how Jaeger represents trace IDs internally.

So I assume your suggestion is still worth discussing since you would be locked into Jaeger as OpenTracing implementation either way and it would make your use case easier to handle.

Marfusios:
Thanks for the info, locking to Jaeger is fine for us. 
I believe I can use `SpanContext.ContextFromString(string)`, [Linkerd sends](https://discourse.linkerd.io/t/logging-to-zipkin/672) coordinates (traceId, spanId) in the header. 
I think my solution is insufficient because of missing parent spanID. 

Btw I found out that you groundbreaking refactored code and I still use version 0.0.9. 
So I will wait for the new released version and then come back to this feature.

Falco20019:
You can subscribe to #62 to be informed once 0.0.10 is released. We plan to release it the next days, since OpenTracing 0.12.0 should also release end of this week.

Issue #57, opened 2018-04-21
Jaeger and the `Microsoft.Extensions.*` stack (DI, Configuration, Logging)


cwe1ss:
ASP.NET Core services are a one of the main targets of this library so it should be as easy as possible to set up a tracer there. We should therefore support the following modules/libraries:

* `Microsoft.Extensions.Configuration`: A developer should be able to configure Jaeger from environment variables, config files, etc. - The `Microsoft.Extensions.Configuration` library supports many different input sources.
* `Microsoft.Extensions.DependencyInjection`: A developer should be able to call `services.AddJaeger()`.
* `Microsoft.Extensions.Logging`: We already have a dependency on that.

With our already existing dependency on `Microsoft.Extensions.Logging` we are currently in a weird limbo state: People who do not use the new `Microsoft.Extensions.*` stack (e.g. ASP.NET WebForms) inherit many `Microsoft.Extensions.*` dependencies and can't easily consume the logs; people who already use the new `Microsoft.Extensions.*` stack (ASP.NET Core) get proper logging support but they don't get DI & Configuration support.

So the first thing I would like to discuss is if we should build this into the library (by adding additional dependencies) or if we want to move all `Microsoft.Extensions.*` stuff into a separate library:

### Scenario 1: Making the library `Microsoft.Extensions.*`-native

Adding additional dependencies would make it a `Microsoft.Extensions`-"native" library like e.g. Entity Framework Core, Identity Server. This would make it very easy to use Jaeger with ASP.NET Core (and any hosting framework that supports that stack) as they only have to add the main `Jaeger` package and call `services.AddJaeger()` and everything else will pretty much work automatically. 

But this obviously also limits its usage possibilities as it makes it much harder to use it in other scenarios.

### Scenario 2: Adding a new `Jaeger.Microsoft.Extensions` library

Thanks to the `Builder` types, we already have a decent initialization story, so it's already quite easy to use the library without dependencies on `Microsoft.Extensions.*`. One thing we don't yet have though (not even with #55 ) is Javas [Configuration.java](https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/Configuration.java) type that supports configuration via environment variables. I did not port this because I wanted to have this discussion first.

[...]

PR #61, opened 2018-05-20
Depend on Logging.Abstractions only (#57)

cwe1ss:
As discussed in #57, we should only depend on the `Microsoft.Extensions.Logging.Abstractions` package in the core Jaeger library.

### DIFF
--- a/src/Jaeger/Jaeger.csproj
+++ b/src/Jaeger/Jaeger.csproj
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.2" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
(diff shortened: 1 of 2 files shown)


src/Jaeger/Jaeger.csproj
```diff
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.0.0" />
```
Falco20019: Why are you using `2.0.0` for `Microsoft.Extensions.Logging.Abstractions` and `2.0.2` for `Microsoft.Extensions.Logging`? `Microsoft.Extensions.Logging` has it's dependency set to `Microsoft.Extensions.Logging.Abstractions` version `2.0.2`, so we should keep those aligned.

cwe1ss: Well, it was intentional because I thought that using 2.0.0 would make it easier for people who aren't on the latest patch version yet to use this but as people should move to the latest patch release asap anyway, there's probably not much point in that.

It now targets 2.0.2 - which is the latest patch release.

PR #63, opened 2018-05-20
Update to OpenTracing 0.12.0

cwe1ss:
This currently targets OpenTracing v0.12.0-rc1. I'll update this PR as soon as the final version is released.

The example currently throws a runtime exception because there's no compatible release of OpenTracing.Contrib.NetCore yet but I also already have an open PR for that.


cwe1ss:
I updated the PR to target the latest versions of OpenTracing & OpenTracing.Contrib.NetCore. 

@Falco20019 Would you have time to review the PR?

Falco20019:
Will do the review tomorrow!

PR #69, opened 2018-05-28
Fixed logging of active span

### LINKED ISSUE #68: Self referencing loop when using LoggingReporter
With the dropping of [IJaegerCoreSpan](https://github.com/jaegertracing/jaeger-client-csharp/pull/55/files#diff-5cf05b611581a56a5de3cb4d55a24902L17) in #55, the `JsonIgnore` attribute from `Tracer` was removed. This is needed to resolve a referencing loop in `JsonConvert` used by `LoggingReporter`. Right now, logging throws this exception:

**Newtonsoft.Json.JsonSerializationException:** 'Self referencing loop detected for property 'Span' with type 'Jaeger.Span'. Path 'Tracer.ScopeManager.Active'.'

This happens, when we are logging the active Span when finishing a scope from `StartActive`. We have two options to resolve this.

This can be easily fixed by adding the `JsonIgnore` attribute to the `Tracer` property in the `Span` implementation.

Falco20019:
Fixes #68 

I added the missing `JsonIgnore` attribute directly. I first wanted to do it through a through an `ContractResolver` but later saw that it's only defined in the implementation and not by the interface in `OpenTracing`. So just ignore the first commit, it will be squashed together before merging.

I also added a test case for regression.


cwe1ss:
see #68 - I don't think this is a good approach as it exposes too much and requires making stuff public that maybe shouldn't be. I think we should just follow Java's example and use span.ToString() which outputs details about the context and the operation name. 
What's your use case where you need all/more details?

Falco20019:
I will adjust the `LoggingReporter` to just use `span.ToString()` then. I think we also can drop the `Newtonsoft.Json` dependency then. We'll see.

PR #71, opened 2018-05-28
Ported ToString() methods from java

Falco20019:
The [Configuration.GetTracer()](https://github.com/jaegertracing/jaeger-client-java/blob/3ffaf9d753fe96a1925729d42fd371a0a166c03b/jaeger-core/src/main/java/io/jaegertracing/Configuration.java#L215) method is getting a printable representation of the Tracer and all it's referenced classes. The java implementation uses `lombok.ToString` annotation for this ([see Tracer](https://github.com/jaegertracing/jaeger-client-java/blob/3ffaf9d753fe96a1925729d42fd371a0a166c03b/jaeger-core/src/main/java/io/jaegertracing/Tracer.java#L57)). In C# we have to implement the `ToString()` methods ourselves.

This highly improves readability of the debug log. Right now, it only says `Initialized Jaeger.Tracer` for each of my tracers. I would have wanted to have at least the `ServiceName` printed.

### DIFF
--- a/src/Jaeger.Thrift/Senders/Internal/ThriftUdpClientTransport.cs
+++ b/src/Jaeger.Thrift/Senders/Internal/ThriftUdpClientTransport.cs
@@ -105,5 +105,10 @@ namespace Jaeger.Thrift.Senders.Internal
             }
             _isDisposed = true;
         }
+
+        public override string ToString()
+        {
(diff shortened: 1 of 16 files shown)


Falco20019:
This is the last issue I see as mandatory for the [v0.1.0 milestone](https://github.com/jaegertracing/jaeger-client-csharp/milestone/1). After that, I will make sure that the documentation and tutorials are matching that version so that we have a good and strong pre-release.

PR #79, opened 2018-05-29
Updated README files to show current state

### LINKED ISSUE #37: Update README to match current code
The README currently does not go along with all the code changes that happened in the last couple of weeks. Lets get that fixed.

Falco20019:
Fixes #37 

The README.md was outdated. I appended some more data, also on `Configuration`. I will also make sure to update the `Sampler` README and add a `Reporter` README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,17 +12,18 @@ This library is still under construction and needs to be peer reviewed as well a
 This package contains everything you need to get up and running. If you want to report to a system such as Jaeger or Zipkin you will need to use their NuGet packages.
 
 ### The Tracer
-The following will give you a tracer that reports spans to the `ILogger` from `ILoggerFactory`.
+The following will give you a tracer that reports spans to an `ILogger` instance from `ILoggerFactory`.
 
 ```C#
 using Jaeger;
 using Jaeger.Reporters;
 using Jaeger.Samplers;
+using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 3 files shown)


Falco20019:
Do we want to link to the tutorial? It's currently at https://github.com/saschaholesch/opentracing-tutorial/tree/csharp/csharp and not merged yet. I would still already create the link to where it would be after the merge. I want to get it merged by tomorrow.

Falco20019:
I will release `0.1.0` once this is done, since I want to have the changes included in the tag. #44 does not change anything in this repository, so it should be fine to get it done after the release, since it's a bit of a circular dependency. (Tutorial targets latest 0.1.0 which is not out, we have #44 as milestone target which would block release.)

yurishkuro:
I'd recommend merging the tutorial anyway, with a WIP caption if necessary.

Falco20019:
@yurishkuro I will do a pre-release now so that reviewing the tutorial can be done with the pre-release. Only thing missing are the READMEs that I want to have in the tag. So technically, the `nupkg` should stay the same.

Falco20019:
@yurishkuro I added you for review since this does not need C# or jaeger-client-csharp knowledge. I copy-pasted every code snippet from tests yesterday, so they have to be correct. So it should be fine to just check the README files for logical errors or misleading parts.

Issue #80, opened 2018-06-01
Required field Process is not set


Marfusios:
Hello, 

after trying the latest version (0.1.0), I get an error from Jaeger server: 
`Unable to process request body: Required field Process is not set\n`. 
Could you help me? 

My configuration: 
```csharp
            var senderConfiguration = new Configuration.SenderConfiguration(loggerFactory)
                .WithEndpoint($"http://{jaegerConfig.ServerIp}:{jaegerConfig.ServerPort}/api/traces");

            var reporterConfiguration = new Configuration.ReporterConfiguration(loggerFactory)
                .WithSender(senderConfiguration);

            var samplerConfiguration = new Configuration.SamplerConfiguration(loggerFactory)
                .WithType("const");

            Tracer = new Configuration(serviceInfo.ServiceId, loggerFactory)
                .WithReporter(reporterConfiguration)
                .WithSampler(samplerConfiguration)
                .GetTracer();
```

Usage:
```csharp
           var builder = _tracer.Tracer
                .BuildSpan("HomeControler request")
                .WithTag("Server", Tags.SpanKindServer)
                .WithTag("Client", Tags.SpanKindClient)
                .WithTag(Tags.DbInstance.Key, "")
                .WithTag(Tags.DbType.Key, "sql");

            using (var scope = builder.StartActive(true))
            {
                var span = scope.Span;

                ViewBag.Title = "My Sample";
                span.SetOperationName("Home page");
                span.SetTag("smth", "special value");

                using (var scope2 = _tracer.Tracer.BuildSpan("Home page 2").StartActive(true))
                {
                    var span2 = scope2.Span;
                    span2.Log("Before scope 2");

                    Task.Delay(2000).Wait();

                    span2.Log("After scope 2");
                }

                span.SetTag("smth2", "special value 2");
            }
```

BTW, a lot of errors were swallowed, so I had to debug it locally to figure out the real issue. 
For example: 
`RemoteReporter`, line 135 (`catch (SenderException ex)`) only report through `_metrics.ReporterFailure`, but not to `_logger`, so when I forgot `http://` in an url, it didn't do anything. 
`THttpClientTransport`, line 180 (`msg.EnsureSuccessStatusCode();`) it throws exception, but doesn't return message from server. 
If you help me with this issue, I can provide PR for logging correction.

Falco20019:
Thanks for reporting. I will have a look soon.

Falco20019:
[...]

PR #82, opened 2018-06-01
Only use 64bit TraceID for compatability

### LINKED ISSUE #81: Correlating traces between csharp and java clients
Im trying to set up a demo to demonstrate tracing across various languages supported by opentracing.

I am running into an issue when trying to get a trace between csharp and java.

I think i have tracked it down to the fact that the traceid generated by the chsarp client is too long.

It seems that the traceid is 32 characters long in the generated header but all the traceids generated by both java and python are only 16 character.

csharp example
e25d900a5c462be7cafa992548dc6d0d

java example
cafa992548dc6d0d

I havent been quite able to get to the bottom of why this might be and what the issue is with the longer id on the java end as it seems to cause an issue and make the java client create a new traceid and breaks the cross process tracing.

I think i have tracked the cause down to the tostring method on TraceID here https://github.com/jaegertracing/jaeger-client-csharp/blob/master/src/Jaeger/TraceId.cs#L38

But im not sure if this is right or wrong.

Falco20019:
Fixes #81 

I added a TODO remark for later fix. I think, the cleanest option would be to have it in `Configuration`. But this would make our `Configuration` and `Tracer` implementations differ from `Java`.

### DIFF
--- a/src/Jaeger/TraceId.cs
+++ b/src/Jaeger/TraceId.cs
@@ -15,7 +15,8 @@ namespace Jaeger
 
         public static TraceId NewUniqueId()
         {
-            return new TraceId(Utils.UniqueId(), Utils.UniqueId());
+            // TODO: Only using 64bit for compatability. Make it configurable to use 128bit TraceIds.
+            return new TraceId(Utils.UniqueId());
         }
 
         public TraceId(long low) : this(0, low)


yurishkuro:
cf. https://github.com/jaegertracing/jaeger/issues/858

Falco20019:
@ScottGuymer yep, just creation now was the problem. When receiving both work and we have unit tests for that.

@yurishkuro Good to have it tracked on meta level. I will merge it once a review form a collaborator was done. I will create another issue to track the second part with configurability.

ScottGuymer:
How does this get pushed out to nuget now it's been merged?

Falco20019:
There is a second issue that should be included in this version (#80). Deadline I gave is Sunday. Do you need it today or is tomorrow early enough? If you need it earlier, I could do a pre-release.

ScottGuymer:
No rush. Probably won't be working on it again until monday anyway.

PR #84, opened 2018-06-04
Fixed HttpSender

### LINKED ISSUE #80: Required field Process is not set

Falco20019:
Fixes #80 

We got this error introduced by `v0.0.10`. I used the implementation from Java again which differed from our port.

### DIFF
--- a/src/Jaeger/Senders/HttpSender.cs
+++ b/src/Jaeger/Senders/HttpSender.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Jaeger.Exceptions;
 using Jaeger.Thrift.Agent;
+using Thrift.Protocols;
 using Thrift.Transports.Client;
 using ThriftBatch = Jaeger.Thrift.Batch;
 using ThriftProcess = Jaeger.Thrift.Process;
@@ -18,7 +19,7 @@ namespace Jaeger.Senders
         private const string HttpCollectorJaegerThriftFormatParam = "format=jaeger.thrift";
         private const int OneMbInBytes = 1048576;
 
-        private readonly Agent.Client _agentClient;
+        private readonly TProtocol _protocol;
         private readonly THttpClientTransport _transport;
 
         /// <summary>
@@ -45,7 +46,7 @@ namespace Jaeger.Senders
             }
 
             _transport = new THttpClientTransport(collectorUri, customHeaders);
-            _agentClient = new Agent.Client(ProtocolFactory.GetProtocol(_transport));
+            _protocol = ProtocolFactory.GetProtocol(_transport);
         }
 
         protected override async Task SendAsync(ThriftProcess process, List<ThriftSpan> spans, CancellationToken cancellationToken)
@@ -53,7 +54,8 @@ namespace Jaeger.Senders
             try
             {
                 var batch = new ThriftBatch(process, spans);
-                await _agentClient.emitBatchAsync(batch, cancellationToken).ConfigureAwait(false);
+                await batch.WriteAsync(_protocol, cancellationToken);
+                await _protocol.Transport.FlushAsync(cancellationToken);
             }
             catch (Exception ex)
             {
--- a/src/Jaeger/Senders/HttpSender.cs
+++ b/src/Jaeger/Senders/HttpSender.cs
@@ -54,8 +54,8 @@ namespace Jaeger.Senders
             try
             {
                 var batch = new ThriftBatch(process, spans);
-                await batch.WriteAsync(_protocol, cancellationToken);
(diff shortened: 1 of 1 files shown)


src/Jaeger/Senders/HttpSender.cs
```diff
@@ -45,15 +46,16 @@ private HttpSender(Builder builder)
...
+            _protocol = ProtocolFactory.GetProtocol(_transport);
         }
 
         protected override async Task SendAsync(ThriftProcess process, List<ThriftSpan> spans, CancellationToken cancellationToken)
         {
             try
             {
                 var batch = new ThriftBatch(process, spans);
-                await _agentClient.emitBatchAsync(batch, cancellationToken).ConfigureAwait(false);
+                await batch.WriteAsync(_protocol, cancellationToken);
```
cwe1ss: .ConfigureAwait(false) is missing on both await calls.

src/Jaeger/Senders/HttpSender.cs
```diff
@@ -45,15 +46,16 @@ private HttpSender(Builder builder)
...
-            _agentClient = new Agent.Client(ProtocolFactory.GetProtocol(_transport));
+            _protocol = ProtocolFactory.GetProtocol(_transport);
         }
 
         protected override async Task SendAsync(ThriftProcess process, List<ThriftSpan> spans, CancellationToken cancellationToken)
         {
             try
             {
                 var batch = new ThriftBatch(process, spans);
-                await _agentClient.emitBatchAsync(batch, cancellationToken).ConfigureAwait(false);
```
cwe1ss: The existing UdpSender uses the same logic and that works fine, doesn't it? Is there different behavior in the UDP/HTTP endpoints of Jaeger?

src/Jaeger/Senders/HttpSender.cs
Falco20019: Will add them in a second. Thanks!

src/Jaeger/Senders/HttpSender.cs
Falco20019: There is a bit of asynchronity here. UDP uses `Agent.EmitBatch` with `TCompactProtocol` to send the `ThriftBatch` and HTTP is directly serializing `ThriftBatch` with `TBinaryProtocol` and sending it as byte array ([see Go endpoint](https://github.com/jaegertracing/jaeger/blob/8ed44f2dd9eeb71e323cd4ff93b1fee146fc6442/cmd/collector/app/http_handler.go#L70)). I made the same mistake when doing the `LetsTrace` HttpSender :)

Issue #99, opened 2018-08-24
Http Thrift requests without Content-Type headers are rejected 


rwkarg:
## Requirement - what kind of business use case are you trying to solve?

Report Spans via Http Thrift requests (port 14268)

## Problem - what in Jaeger blocks you from solving the requirement?

Recent versions of the Jaeger backend reject Http Thrift requests that do not contain a `Content-Type: application/x-thrift` header.
An Http 400 response is returned with the following message:
`Cannot parse content type: mime: no media type`

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Include the expected header in Http Thrift requests

rwkarg:
This was addressed by #101

Falco20019:
Sorry for the delays but I‘m low on time right now :( wanted it as part of v0.3 but just moved it to v0.2.1 which should release the next couple of minutes.

Aaronontheweb:
@Falco20019 thanks! Just upgraded. I really appreciate you all responding so quickly. This solved my problem.

PR #100, opened 2018-08-24
Add Content-Type header to Http Thrift requests

rwkarg:
The current version of the all-in-one container will reject an Http Thrift request (port 14268) if it doesn't have a `Content-Type: application/x-thrift` header.
Older versions would accept those requests without the header (I'm not sure exactly which version the change in behavior started in).

I've tested with the "old" version and it will accept spans either with or without the header.
The latest version will only accept spans with the header. Spans without the header will result in a 400 response with the message `Cannot parse content type: mime: no media type`

Signed-off-by: Ryan Karg <rwkarg@gmail.com>

- Issue #99
- Http Thrift requests sent to the latest Jaeger backend are rejected if they do not specify a `Content-Type=application/x-thrift` header

- Add the expected header

### DIFF
--- a/src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
+++ b/src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
@@ -40,6 +40,7 @@ namespace Thrift.Transports.Client
 
         private bool _isDisposed;
         private MemoryStream _outputStream = new MemoryStream();
+        private static readonly MediaTypeWithQualityHeaderValue XThriftMediaType = new MediaTypeWithQualityHeaderValue("application/x-thrift");
 
         public THttpClientTransport(Uri u, IDictionary<string, string> customHeaders)
             : this(u, Enumerable.Empty<X509Certificate>(), customHeaders)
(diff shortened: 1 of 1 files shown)


src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
```diff
@@ -40,6 +40,7 @@ public class THttpClientTransport : TClientTransport
 
         private bool _isDisposed;
         private MemoryStream _outputStream = new MemoryStream();
+        private static readonly MediaTypeWithQualityHeaderValue XThriftMediaType = new MediaTypeWithQualityHeaderValue("application/x-thrift");
```
yurishkuro: Thanks for the fix. While x-thrift is accepted, it is better to be more specific about which encoding it is, i.e. ""application/vnd.apache.thrift.binary""

https://github.com/jaegertracing/jaeger/blob/7262bdc9e59933f15fb0802f0d3ba7fb642a8955/cmd/collector/app/http_handler.go#L39

rwkarg:
hmm.. I followed the example for sign-off in the Contributing Guidelines:
`Signed-off-by: Bender Bending Rodriguez <bender.is.great@gmail.com>`
which matches what I used for a previous PR (#91):
`Signed-off-by: Ryan Karg <rwkarg@gmail.com>`

Has the required format changed?

rwkarg:
Closing this and creating #101 to address DCO failure

PR #101, opened 2018-08-24
Add Content-Type header to Http Thrift requests

rwkarg:
The current version of the all-in-one container will reject an Http Thrift request (port 14268) if it doesn't have a `Content-Type` header.
Older versions would accept those requests without the header (I'm not sure exactly which version the change in behavior started in).

I've tested with the "old" version and it will accept spans either with or without the header.
The latest version will only accept spans with the header. Spans without the header will result in a 400 response with the message `Cannot parse content type: mime: no media type`

- Issue #99
- Http Thrift requests sent to the latest Jaeger backend are rejected if they do not specify a `Content-Type` header

- Add the expected header

### DIFF
--- a/src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
+++ b/src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
@@ -40,6 +40,8 @@ namespace Thrift.Transports.Client
 
         private bool _isDisposed;
         private MemoryStream _outputStream = new MemoryStream();
+        private static readonly MediaTypeWithQualityHeaderValue XThriftMediaType = new MediaTypeWithQualityHeaderValue("application/x-thrift");
+        private static readonly MediaTypeWithQualityHeaderValue ApacheThriftMediaType = new MediaTypeWithQualityHeaderValue("application/vnd.apache.thrift.binary");
 
         public THttpClientTransport(Uri u, IDictionary<string, string> customHeaders)
             : this(u, Enumerable.Empty<X509Certificate>(), customHeaders)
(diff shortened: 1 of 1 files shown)


rwkarg:
I believe that the appveyor/branch check will not start/pass since this is a PR from a fork as opposed to a branch.

src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
```diff
@@ -148,7 +150,7 @@ private HttpClient CreateClient()
                 httpClient.Timeout = TimeSpan.FromSeconds(_connectTimeout);
             }
 
-            httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/x-thrift"));
+            httpClient.DefaultRequestHeaders.Accept.Add(XThriftMediaType);
```
yurishkuro: please elaborate - what is the reason for using a different media type here?

rwkarg: That was existing (just pulled out to a variable). If the Accept header should also be updated to the Apache Thrift type then I can do that, too.

rwkarg: Looking at the Java client, it doesn't look like it's setting the Accepts header. Is that even necessary at all?

rwkarg: Looks like this is how the apache/thrift source this was pulled from is set up
https://github.com/apache/thrift/blob/master/lib/netcore/Thrift/Transports/Client/THttpClientTransport.cs#L151

Falco20019: Can you try to set it to ApacheThriftMediaType? This sets the accept header and I assume it works. Can’t test it myself before Monday.

PR #106, opened 2018-09-07
Add client certificates only if any provided

### LINKED ISSUE #105: [Thrift] HttpClientHandler.ClientCertificates not present in some .NET Framework versions

Report tracing from Windows machines running various .NET Framework versions

The THttpClientTransport class always attempts to add additional client certificates to the underlying HttpClientHandler. That property access a class that is part of the netstandard 2.0 API surface, but not implemented on certain .NET Framework version. The result is a MissingMethodException that prevents the reporter from being created.

Adding a conditional check around adding additional ClientCertificates to only execute if there are additional certificates provided by the consumer of the library.

I had considered catching and ignoring the MissingMethodException but that would result in silently dropping the additional certificates that the consumer expects to be added. On an older .NET Framework where that method is not available then it would be preferable to have an exception thrown if the intended behavior is not achievable.
There is a THttpClientTransport constructor that doesn't take in additional certificates which would allow this method to continue to run as intended by bypassing the ClientCertificates call.
In actuality, HttpSender doesn't even expose a way to pass in additional certificates so the only way I can see this used is the path that doesn't have any certificates provided.

rwkarg:
Not all .NET Framework versions that claim to be netstandard2.0 
compliant implement methods necessary for the 
ClientCertificates property to be accessed and result in a 
MissingMethodException being thrown. 
This conditional will only attempt to access that property if 
additional client certificates are provided. 
Additionally, the only usage by HttpSender to create a 
THttpClientTransport does not ever pass in additional certificates 
so this will always skip over the property access. 
As a result, this library can be used on older .NET Frameworks that 
otherwise implement the netstandard2.0 framework.

- Resolves #105 

- Conditionally access the ClientCertificates property only if additional client certificates are provided

### DIFF
--- a/src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
+++ b/src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
@@ -140,7 +140,8 @@ namespace Thrift.Transports.Client
         private HttpClient CreateClient()
         {
             var handler = new HttpClientHandler();
-            handler.ClientCertificates.AddRange(_certificates);
+            if(_certificates.Length > 0)
+                handler.ClientCertificates.AddRange(_certificates);
 
             var httpClient = new HttpClient(handler);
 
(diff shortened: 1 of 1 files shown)


Falco20019:
Sorry, I‘m a bit low on time right now. So reviews are delayed a bit currently.

PR #108, opened 2018-10-05
Refactor method for resolving configuration properties (#57)

thzinc:
_I welcome feedback on this work. Thanks for your time in reviewing!_

- #57 

Updated the method signatures that build configuration `FromEnv` to take an optional PropertyResolver delegate that can be used to override the default behavior of resolving properties from environment variables.

This is intended to be an alternative solution to directly depending on Microsoft.Extensions.Configuration or similar sources of configuration data.

### DIFF
--- a/src/Jaeger/Configuration.cs
+++ b/src/Jaeger/Configuration.cs
@@ -152,18 +152,21 @@ namespace Jaeger
         }
 
         /// <summary>
-        /// Returns <see cref="Configuration"/> object from environmental variables.
+        /// Returns <see cref="Configuration"/> object from the property resolver
         /// </summary>
-        public static Configuration FromEnv(ILoggerFactory loggerFactory)
+        /// <param name="loggerFactory"></param>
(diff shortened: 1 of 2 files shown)


Falco20019:
Thanks for your input. This would target at least one part of the problem of #57. We still have Logging and DependencyInjection open, but this could be a starting point. For one, we waited for what Microsofts plans to do in those cases with future frameworks, on the other side, also no one had really time to tackel them yet.

But I think short-term having a way to use Configuration class with some frameworks like `Microsoft.Extensions.Configuration` instead of only `Environment` would be a nice improvement. Even if it just opens the API and not fully features easy-to-use packages like with `Microsoft.Extensions.Logging.Absractions`  yet.

I will give it a look, hopefully the next days.

Falco20019:
I think #110 should be fine to use. Everyone can add his own {{IConfiguration}} implementation and it will reduce having to use an additional delegate on top of {{IConfiguration}}. Is this OK for you?

thzinc:
@Falco20019, I prefer the solution in #110 using `IConfiguration`. I'm happy to close this PR in deference to #110.

PR #109, opened 2018-10-12
Add commented-out Jaeger PackageReference

enesunal:
First time learners need to break dependency from example project to parent project. 

I know binding example application directly to the current available version is more appropriate; however, giving an example to use NuGet package could be useful.

### DIFF
--- a/examples/Jaeger.Example.WebApi/Jaeger.Example.WebApi.csproj
+++ b/examples/Jaeger.Example.WebApi/Jaeger.Example.WebApi.csproj
@@ -7,6 +7,12 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.5" />
     <PackageReference Include="OpenTracing.Contrib.NetCore" Version="0.4.0" />
+
+    <!-- NOTE:
+        In case you're building this application standalone, use below PackageReference 
+        with appropriate an appropriate version instead of relative-path ProjectReference 
+        to Jaeger project -->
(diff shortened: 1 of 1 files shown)


examples/Jaeger.Example.WebApi/Jaeger.Example.WebApi.csproj
```diff
@@ -7,6 +7,12 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.5" />
     <PackageReference Include="OpenTracing.Contrib.NetCore" Version="0.4.0" />
+
+    <!-- NOTE:
+        In case you're building this application standalone, use below PackageReference 
+        with appropriate an appropriate version instead of relative-path ProjectReference
```
Falco20019: "with [-appropriate] an appropriate version " There is a second "appropriate" that needs to be removed. Other than that, I'm fine with the suggestion.

A more detailed tutorial can be found at https://github.com/yurishkuro/opentracing-tutorial/tree/master/csharp Not sure if it is linked somewhere, if not, we also may should add a link in the README.md to it.

enesunal: @Falco20019 I've fixed it now. At first sight, I couldn't see if there is any link to tutorial repo.

enesunal:
@Falco20019 ready to go.

Falco20019:
Thanks!

PR #110, opened 2018-10-12
Enrich the ways one can configure Jaeger with ASP.Net's IConfiguration

mihaipopescu:
- Have multiple ways in which Jaeger can be _easily_ configured, not only through Environment Variables.

- Sometimes it's very useful to use other configuration providers and IConfiguration offers this possibility.

see https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.1


mihaipopescu:
If the direction is accepted, I think some documentation changes will be nice to add as well.

thzinc:
I prefer this solution over the compromise I proposed in #108. 👍

PR #111, opened 2018-10-17
Updated NuGET packages

Falco20019:
- Microsoft.AspNetCore.All < 2.0.9 has some security issues

- Updated all non-thrift packages to current versions.

@cwe1ss Is it problematic to update `Microsoft.Extensions.Logging.Abstractions` or any other package in our libraries?

### DIFF
--- a/Jaeger.sln
+++ b/Jaeger.sln
@@ -1,4 +1,5 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.27130.2024
 MinimumVisualStudioVersion = 15.0.26124.0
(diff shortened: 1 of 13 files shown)


cwe1ss:
.NET Core 2.0 has recently [reached end of life](https://github.com/dotnet/announcements/issues/86) so I guess it should be fine to update the logging library to `2.1.x`.

cwe1ss:
So you could also update all examples to .NET Core 2.1 if you want?!

Falco20019:
Good idea. Didn’t think about the EOL for 2.0 :)

Falco20019:
@cwe1ss Have updated and upgraded everything to netcoreapp2.1 and solved some warnings.

Falco20019:
I removed the examples from build to avoid a problem with `Microsoft.WebApplication.targets` in `Jaeger.Example.WebApi.NetFx`. `Jaeger.Example.WebApi` would work but I think we should also keep it out since we don't really generate artefacts from it.

Falco20019:
@cwe1ss Friendly ping :)

Falco20019:
@cwe1ss @grounded042 Would be great if one of you could have a quick look and approve the review.

PR #117, opened 2018-12-28
Use System.Threading.Channels as RemoteReporter's internal queue

csyszf:
Signed-off-by: Sun Zhongfeng <csyszf@outlook.com>

- Increase queue throughput and reduce dropped spans

- Replace RemoteReporter's internal queue by Channel<T>

I wrote a simple benchmark for it, and saw a significant throughput increase(50% - 150% on my laptop).


rwkarg:
In testing, I actually have a lot more dropped spans from the Channel implementation. Would it make more sense to have a ReportAndWait method on a Reporter that would block (or actually return a Task that can be awaited) to measure actual throughput? We're already driving this much harder than it could ever go in real usage since the sender is virtually a noop.

```
reporter: ChannelRemoteReporter
try report 10000000 spans
send 2603463 spans in 5524 ms
throughput: 471k/s

reporter: BlockingCollectionRemoteReporter
try report 10000000 spans
send 9897254 spans in 7979 ms
throughput: 1240k/s
```

rwkarg:
DataFlow was simpler to implement, slightly better (fewer) dropped metrics, but slower.

The channel implementation was best at not dropping metrics and was faster:

```
reporter: DataFlowRemoteReporter
try report 10000000 spans
send 9939308 spans in 9168 ms
throughput: 1084k/s

reporter: ChannelRemoteReporter
try report 10000000 spans
send 9967039 spans in 6829 ms
throughput: 1459k/s

reporter: BlockingCollectionRemoteReporter
try report 10000000 spans
send 9913289 spans in 8165 ms
throughput: 1214k/s
```

csyszf:
@rwkarg DataFlow's a good option, but IMO it's more about "processing" than "queueing", it's the data-flow library but there's few data-process logic in reporter layer, it will be more useful if we adapt it to the full report-send flow, and Channels/ConcurrentQueue seems simpler and more transparent to me .

This micro-benchmark aim to measure the throughput of the queue itself, I can do a real-world benchmark with our test agent later.

And ReportAndWait or "blocked report" is meaningful sometimes - not only in benchmarks, some of our teams they want more "reliable" tracing data (with no dropping), even it will make applications less responsive(wait to report) or bring uncontrolled memory risk(unbounded queue).

benchmarks/Jaeger.Benchmarks/Program.cs
```diff
@@ -6,7 +6,15 @@ class Program
     {
         static void Main(string[] args)
         {
-            BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args);
+            switch (args[0])
```
Falco20019: args.Length could be 0

benchmarks/Jaeger.Benchmarks/Jaeger.Benchmarks.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
Falco20019: What was changed in this line?

src/Jaeger/Jaeger.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
Falco20019: What was changed?

benchmarks/Jaeger.Benchmarks/Program.cs
csyszf: Yeah, fixed.

benchmarks/Jaeger.Benchmarks/Jaeger.Benchmarks.csproj
csyszf: My VS saves *.csproj files as UTF-8 BOM encoding, fixed.

csyszf:
@Falco20019  Sorry for the late response, we can get the ReporterDropped count by calculation if we add reported count and appended span count metrics.

Falco20019:
@csyszf We should also benchmark the effect of having to calculate it as it might affect the throughput.

Issue #116, opened 2018-12-05
Invert dependency between Jaeger and Jaeger.Thrift NuGet packages


avkom:
## Requirement

We do not want that `Jaeger` NuGet package depends on `Jaeger.Thrift` package.

## Problem

`Jaeger` package depends on `Jaeger.Thrift` package. `Jaeger` package may be a 'reference implementation' of OpenTracing standard on .NET platform that can be used with custom implementation of `Jaeger.Senders.ISender` interface. In case our custom implementation of `ISender` does not use Thrift protocol, we do not want that `Jaeger` NuGet package depends on `Jaeger.Thrift` and further `Jaeger.Thrift.VendoredThrift` packages.

## Proposal

1. Invert dependency between `Jaeger` and `Jaeger.Thrift` packages. After this change, `Jaeger.Thrift` package will depend on `Jaeger`.
2. Move implementations of `Jaeger.Senders.ISender` interface from `Jaeger` project to `Jaeger.Thrift` project. (Move classes `HTTPSender`, `ThriftSender`, `UdpSender`, `JaegerThriftSpanConverter`, `Configuration` etc.).
3. After this change, developers have to reference `Jaeger.Thrift` NuGet package in their applications.

Challenges:
1. `RemoteReporter.Builder.Build()` depends on `UdpSender`. We need to break this dependency. For example, raise an exception when sender is not provided.

Falco20019:
I would like to get rid of Thrift once gRPC support is fully added. I also would like to get rid of linking directly to the package as discussed in #76 but currently have little time to work on the library. I hope to find some time in January.

yurishkuro:
Fwiw a similar refactoring happened in the Java client.

avkom:
I am working on a proof of concept for this change, but it requires some polish. I am going to submit a pull request. Do you accept PRs in this repository?

yurishkuro:
> Do you accept PRs in this repository?

certainly

Falco20019:
I think this will somewhat revive the discussion around #34. Maybe we should shift Jaeger back into Jaeger.Core and strip the Transport layer out again. Jaeger could than be the alias to to Jaeger.Thrift as discussed in #34 so that people can just update without breaking.

avkom:
[...]

Issue #76, opened 2018-05-29
Dependency on Thrift


Falco20019:
The follow-up discussion from [here](https://github.com/jaegertracing/jaeger-client-csharp/pull/55#issuecomment-382172354).

The open questions are:
* Do we want to keep our dependency on `Thrift`?

Java thinks about modularization and using `Configuration` and `service load pattern` to only load what's needed. We would have to analyze how to do this in C#. I assume it would be similar to how `Microsoft.Extensions.Logging` and `ASP.NET Core` are abstracting everything into packages and enabling it though adding a middleware/provider and calling an extension method on `Configuration` presumably.

animusblue:
I'm currently having an issue with using this library because of the dependency on the Jaeger.Thrift.VendoredThrift package. It seems that the Jaeger.Thrift.VendoredThrift is a custom copy that doesn't have the asp.net core references. This is causing an issue with our own thrift library as both the nuget package and the default implementation (through a submodule referencing the thrift master) are using the same Thrift namespace. Since Jaeger.Thrift.VendoredThrift isn't a standard copy we can't switch to using it (we need the http protocol). Should the nuget package be placed under it's own namespace to prevent direct conflicts with the master thrift library?

Falco20019:
#116 also has problems with Thrift. Is there finally an up2date nuget package of Thrift? A year ago we only did a request at the issue tracker to finally update it.

animusblue:
We've only found 3rd party NuGet packages and most of them were out of date. We're hesitant to rely on non-official sources for something like this which is why we're using a submodule.

Falco20019:
There finally are official packages available for Thrift at https://www.nuget.org/packages/ApacheThrift/

PR #118, opened 2019-01-14
Replaced Jaeger.Thrift.VendoredThrift with official ApacheThrift

Falco20019:
Related to #76 and #116

We like to get rid of `Jaeger.Thrift.VendoredThrift` and use the new official NuGET package. Right now this is not possible, since it's lacking a strong name.

This PR prepares for this this and is kept WIP until strong naming is enabled for the official package.
https://issues.apache.org/jira/browse/THRIFT-4736


rwkarg:
Just for tracking, the THRIFT issue for signing got duped to https://issues.apache.org/jira/browse/THRIFT-2947

Falco20019:
https://www.nuget.org/packages/ApacheThrift/0.12.0.1 got released with signing. I will check if we can replace the used version `0.11.0` with `0.12.0.1`.

Falco20019:
Thrift 0.12.0.1 does still contain the unsigned DLLs and is therefore not resolving the problem. I created a follow-up issue: https://issues.apache.org/jira/browse/THRIFT-4907

Jens-G:
> Thrift 0.12.0.1 does still contain the unsigned DLLs and is therefore not resolving the problem. I created a follow-up issue: https://issues.apache.org/jira/browse/THRIFT-4907

Incorrect. The DLLs are signed. They just don't have a strong nane. That's not quite the same.

Falco20019:
Regeneration of files for 0.13 is blocked by https://github.com/ahawkins/docker-thrift/pull/22

Falco20019:
Looks like 0.13.0 of ApacheThrift is signed and can be used. Tests are also all green.

PR #126, opened 2019-02-25
Adding identifying info to warning logging

### LINKED ISSUE #125: Warning for already finished Span is not useful

Correctly handling lifetime (with possible disposal) of Spans.

Warning log for already finished spans does not provide any information about which span needs to be fixed.

Add Operation Name and Trace/Span Id to the warning line so that the offending spans can be identified for future correction.

No open questions.

rwkarg:
It is difficult to identify which is the offending span from the current log line. This change is to include information to easily identify which span needs to be addressed; either through a common operation triggers multiple warning log lines or through the specific trace/span id that can be investigated through the dashboard.

- Resolves #125

- Add Operation Name and Trace/Span Id to Warning logging

### DIFF
--- a/src/Jaeger/Span.cs
+++ b/src/Jaeger/Span.cs
@@ -118,7 +118,7 @@ namespace Jaeger
             {
                 if (FinishTimestampUtc != null)
                 {
-                    Tracer.Logger.LogWarning("Span has already been finished; will not be reported again.");
+                    Tracer.Logger.LogWarning("Span has already been finished; will not be reported again. Operation: {0} Trace Id: {1} Span Id: {2}", OperationName, Context.TraceId, Context.SpanId);
                     return;
                 }
                 FinishTimestampUtc = finishTimestampUtc;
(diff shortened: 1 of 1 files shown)


src/Jaeger/Span.cs
```diff
@@ -118,7 +118,7 @@ private void FinishInternal(DateTime finishTimestampUtc)
             {
                 if (FinishTimestampUtc != null)
                 {
-                    Tracer.Logger.LogWarning("Span has already been finished; will not be reported again.");
+                    Tracer.Logger.LogWarning("Span has already been finished; will not be reported again. Operation: {0} Trace Id: {1} Span Id: {2}", OperationName, Context.TraceId, Context.SpanId);
```
Falco20019: Please don‘t use {0}, ... but speaking names. This is not string.Format syntax but structured logging of ILogger. The logging framework will create key-value-pairs for those fields in addition to a stringified version. So just use {operationName} and so on.

Falco20019:
Hi @rwkarg, thanks for the PR! I will merge it right after the keys are fixed.

PR #128, opened 2019-02-27
Open configuration API

### LINKED ISSUE #124: Add public read-only properties to the tracer builder

I'm working on a workaround for #121 where I create a decorator for the configured sampler and set that up during app startup. My decorator can make some sampling decisions (or add some logging, or whatever) and I can still use environment-based configuration for the default sampler types (`const` or whatever).

I envision the setup of the tracer looking like this at app startup:

```c#
{
  var logFactory = serviceProvider.GetRequiredService<ILoggerFactory>();
  var config = Tracing.Configuration.FromEnv(logFactory);
  var tracerBuilder = config.GetTracerBuilder();
  // TODO: Make use of the environment-configured settings, just wrap the sampler.
  var accessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
  var decorator = new SamplerDecorator(tracerBuilder.Sampler, accessor);
  tracerBuilder.WithSampler(decorator);
  var tracer = tracerBuilder.Build();
  GlobalTracer.Register(tracer);
  return tracer;
})
```

The [spot in `Configuration` that generates the `ISampler`](https://github.com/jaegertracing/jaeger-client-csharp/blob/16fbedf556661d1c79d2692c2266f681a02c40c1/src/Jaeger/Configuration.cs#L204) actually does a lot of work tying metrics to the sampler. After that, [`Tracer.Builder.WithSampler` sets a private field](https://github.com/jaegertracing/jaeger-client-csharp/blob/16fbedf556661d1c79d2692c2266f681a02c40c1/src/Jaeger/Tracer.cs#L244) and you can never access that built/wired sampler again.

You also can't re-create the mechanism manually. You can read the configuration from the environment again, but [the `Configuration.SamplerConfiguration.CreateSampler` method is internal](https://github.com/jaegertracing/jaeger-client-csharp/blob/16fbedf556661d1c79d2692c2266f681a02c40c1/src/Jaeger/Configuration.cs#L346) so you can't actually rebuild the sampler from config.

The actual `SamplerConfiguration` object exposes read-only properties with the values that are set by the builder methods (e.g., `WithParam()`). I'd like to see public read-only properties on `Tracer.Builder` that have the values set by the builder methods.

It may also be nice to have properties on the actual `Tracer` to expose the things that have been set and possibly allow post-build configuration. I could just as easily have done something like...

```c#
{
  var logFactory = serviceProvider.GetRequiredService<ILoggerFactory>();
  var config = Tracing.Configuration.FromEnv(logFactory);
  var tracer = config.GetTracerBuilder().Build();
  var accessor = serviceProvider.GetRequiredService<IHttpContextAccessor>();
  var decorator = new SamplerDecorator(tracer.Sampler, accessor);
  tracer.Sampler = decorator;
  GlobalTracer.Register(tracer);
  return tracer;
})
```

This likely relates to #74 which asks about whether there should be constructors or only builders.

Falco20019:
- Resolves #124 

- Read-only properties for all fields configurable through `With...`
- `XConfigration.GetX` methods marked as virtual for overriding

### DIFF
--- a/src/Jaeger/Configuration.cs
+++ b/src/Jaeger/Configuration.cs
@@ -286,7 +286,7 @@ namespace Jaeger
             return this;
         }
 
-        [Obsolete]
+        [Obsolete("Use the property 'TracerTags' instead.")]
         public IReadOnlyDictionary<string, string> GetTracerTags()
         {
             return _tracerTags;
--- a/src/Jaeger/Configuration.cs
+++ b/src/Jaeger/Configuration.cs
@@ -129,13 +129,12 @@ namespace Jaeger
         /// <summary>
         /// The serviceName that the tracer will use.
         /// </summary>
-        private readonly string _serviceName;
+        public string ServiceName { get; }
+
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
-        private CodecConfiguration _codecConfig;
         private IMetricsFactory _metricsFactory;
         private Dictionary<string, string> _tracerTags;
-        private bool _useTraceId128Bit;
 
         /// <summary>
         /// Lazy singleton <see cref="Tracer"/> initialized in <see cref="GetTracer()"/> method.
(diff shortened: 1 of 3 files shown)


Falco20019:
Once this is reviewed and merged, I will release `v0.3.0`. I moved everything related to the `Thrift` change and the topics with ongoing discussions into `v0.4.0`.

src/Jaeger/Configuration.cs
```diff
@@ -284,6 +286,7 @@ public Configuration WithTracerTags(Dictionary<string, string> tracerTags)
             return this;
         }
 
+        [Obsolete]
```
saschaholesch: Add note to use property.

Falco20019: Thanks for noticing.

PR #130, opened 2019-03-06
Made Span.GetTags and Span.GetLogs immutable

Falco20019:
Fixes #129 

- `Span.GetTags()` and `Span.GetLogs()` could be modified while iterating over them.
- Creating copies of the list when necessary. `Span.GetReferences()` and most other locations are not affected since they can't be modified by public methods.

### DIFF
--- a/src/Jaeger/Span.cs
+++ b/src/Jaeger/Span.cs
@@ -10,6 +10,7 @@ namespace Jaeger
     public class Span : ISpan
     {
         private static readonly IReadOnlyList<LogData> EmptyLogs = new List<LogData>().AsReadOnly();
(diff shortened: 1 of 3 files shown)


src/Jaeger/Span.cs
```diff
@@ -53,7 +54,7 @@ public class Span : ISpan
 
         public IReadOnlyList<Reference> GetReferences() => _references;
 
-        public IReadOnlyDictionary<string, object> GetTags() => _tags;
+        public IReadOnlyDictionary<string, object> GetTags() => new Dictionary<string, object>(_tags);
```
yurishkuro: should there be locking?

Falco20019: Added `lock` and `EmptyTags`.

PR #141, opened 2019-04-02
Add GrpcSender for jaeger-collector

### LINKED ISSUE #139: Support for GrpcSender

The new `model.proto` was defined in https://github.com/jaegertracing/jaeger/issues/904 and is accepted on GRPC port `14250` by the collector.

Currently not implemented in the client library.

Falco20019:
-  Fixes #139

- The jaeger-collector got an GRPC endpoint. This can be used instead of `Thrift`. This add a separate package `Jaeger.Grpc` (like `Jaeger.Thrift`).

As discussed in #116, we should invert the dependecy before releasing 0.4.0 so that the user can choose between `Thrift` or `Grpc` and does not have to get all the dependencies (or non in case they implement `ISender` themselves). But this is outside the scope of this PR.

### DIFF
--- a/src/Jaeger/Configuration.cs
+++ b/src/Jaeger/Configuration.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Threading;
 using System.Threading.Tasks;
+using Grpc.Core;
 using Jaeger.Metrics;
 using Jaeger.Propagation;
 using Jaeger.Reporters;
@@ -56,6 +57,21 @@ namespace Jaeger
         /// The port used to locate the agent.
         /// </summary>
         public const string JaegerAgentPort = JaegerPrefix + "AGENT_PORT";
+
+        /// <summary>
+        /// The host name used to locate the GRPC collector.
+        /// </summary>
+        public const string JaegerGrpcHost = JaegerPrefix + "GRPC_HOST";
+
+        /// <summary>
+        /// The port used to locate the GRPC collector.
+        /// </summary>
+        public const string JaegerGrpcPort = JaegerPrefix + "GRPC_PORT";
+
+        /// <summary>
+        /// The root certificate used to create a trusted TLS connection to the GRPC collector.
+        /// </summary>
+        public const string JaegerGrpcRootCertificate = JaegerPrefix + "GRPC_ROOT_CERTIFICATE";
 
         /// <summary>
         /// Whether the reporter should log the spans.
@@ -577,7 +593,7 @@ namespace Jaeger
         }
 
         /// <summary>
-        /// Holds the configuration related to the sender. A sender can be a <see cref="HttpSender"/> or <see cref="UdpSender"/>.
+        /// Holds the configuration related to the sender. A sender can be a <see cref="HttpSender"/>, <see cref="UdpSender"/> or <see cref="GrpcSender"/>.
         /// </summary>
         public class SenderConfiguration
         {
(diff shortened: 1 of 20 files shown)


src/Jaeger/Configuration.cs
```diff
@@ -674,29 +723,65 @@ public virtual ISender GetSender()
...
-                        httpSenderBuilder.WithAuth(AuthUsername, AuthPassword);
-                    }
-                    else if (!string.IsNullOrEmpty(AuthToken))
-                    {
-                        _logger.LogDebug("Auth Token environment variable found.");
-                        httpSenderBuilder.WithAuth(AuthToken);
-                    }
+                    return GetHttpSender();
+                }
+
```
saschaholesch: Remove unnecessary spacing.

Falco20019:
@yurishkuro Looking at the `collector.proto`, I assume we will have a `agent.proto` in phase 3 of https://github.com/jaegertracing/jaeger/issues/773. My proposal is, to implement the `jager-collector` GRPC endpoint like we have with the HTTP endpoint for users skipping the `jaeger-agent`.

I would also propose to add a new environment flag `JAEGER_GRPC_MODE` to define if contacting `jaeger-collector` or `jaeger-agent`. As we only have collector mode right now, I would also spare it for now and just keep it as concept.

src/Jaeger/Configuration.cs
```diff
@@ -668,35 +717,69 @@ public SenderConfiguration WithAuthPassword(string password)
...
+                _logger.LogDebug("Using the HTTP Sender to send spans directly to the endpoint.");
+                return httpSenderBuilder.Build();
+            }
+
+            private ISender GetGrpcSender()
+            {
+                ChannelCredentials credentials;
+                if (!string.IsNullOrEmpty(GrpcRootCertificate))
+                {
+                    _logger.LogDebug("Use TLS GRPC channel with data from the environment variables.");
```
yurishkuro: Use or Using?

src/Jaeger/Jaeger.csproj
```diff
@@ -7,7 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Jaeger.Thrift\Jaeger.Thrift.csproj" />
+      <ProjectReference Include="..\Jaeger.Grpc\Jaeger.Grpc.csproj" />
+      <ProjectReference Include="..\Jaeger.Thrift\Jaeger.Thrift.csproj" />
```
yurishkuro: Doesn't have to be part of this PR, but if one of your goal was "not to have thrift", doesn't it still get linked?

src/Jaeger/Senders/GrpcSender.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    ServiceName = span.Tracer.ServiceName,
+                    Tags = { JaegerGrpcSpanConverter.BuildTags(span.Tracer.Tags) }
+                };
+            }
+
+            GrpcSpan grpcSpan = JaegerGrpcSpanConverter.ConvertSpan(span);
+            _spanBuffer.Add(grpcSpan);
+
+            return Task.FromResult(0);
```
yurishkuro: How is the volume / message size controlled?

src/Jaeger/Jaeger.csproj
Falco20019: Will be part of a separate PR specifically targetting #116 and done before I will release 0.4.0. So this state won't end up in any deployment.

src/Jaeger/Senders/GrpcSender.cs
Falco20019: Currently only by the remote reporters flush timeout. But I can add a controller like we did it with Thrift if you would prefer that.

yurishkuro: I think you always want upper bound on both time and size, otherwise you might accumulate too much data for a single message.

Falco20019:
This needs to be rebased once #151 was merged.

Falco20019:
As #151 should be merged today or tomorrow, I will recreate this PR in a follow-up since it's easier as rebasing this.

src/Jaeger/Senders/GrpcSender.cs
Falco20019: This has been addressed in #169, so I will resolve it here.

src/Jaeger/Configuration.cs
Falco20019: Using :) Addressed in #169

PR #144, opened 2019-04-18
Added support for 128bit to B3 decoder

### LINKED ISSUE #143: 128 bit trace id not working with b3 propagation
> Just added configurability in [jaegertracing/jaeger-client-csharp#94](https://github.com/jaegertracing/jaeger-client-csharp/pull/94) by proposing the following changes to the API:
> 
> * Added environment variable `JAEGER_TRACEID_128BIT`
> * Added `WithTraceId128Bit` method to `Configuration`
> * Added `WithTraceId128Bit` method to `Tracer.Builder`
> * Adde `UseTraceId128Bit` property to `Tracer`
> 
> `Go` and `C++` have done it through their `options` context and do not offer it through `Configuration` (either code or environment) yet. It would be great to discuss if we like to have this in configuration and if, what env-var to use.

Hey I am using "JAEGER_TRACEID_128BIT" environment variable to enable 128 bit trace id support but while I deploy application image in windows container it works fine but when I deploy the same image in kubernetes it never generates 128 bit traceid for the same image with same variables passed for deployement.

**Below is my JAEGER client configuration of c# :**

```
           services.AddOpenTracing();

            // Adds the JAEGER Tracer.
            {
                Tracer tracer = null;
                Jaeger.Configuration jaegerConfig = null;
                var hostingEnv = serviceProvider.GetRequiredService<IHostingEnvironment>();

                // Add project name as service name as a default to JAEGER client if not passed from outside as a parameter.
                serviceName = serviceName ?? hostingEnv.ApplicationName;
                try
                {
                    jaegerConfig = hostingEnv.IsDevelopment()
                        ? new Jaeger.Configuration(serviceName, loggerFactory)
                        : Jaeger.Configuration.FromEnv(loggerFactory);
                }
                catch (Exception)
                {
                    jaegerConfig = new Jaeger.Configuration(serviceName, loggerFactory);
                }

                // Get tracer instance with configurations.
                tracer = (Tracer)jaegerConfig
                    .WithSampler(Jaeger.Configuration.SamplerConfiguration.FromEnv(loggerFactory)) // This will fetch all the configurations for Sampler from environment variables.
                    .WithReporter(Jaeger.Configuration.ReporterConfiguration.FromEnv(loggerFactory)) // This will fetch all the configurations for Reporter from environment variables.
                    .GetTracer();

                // Allows code that can't use DI to also access the tracer.
                GlobalTracer.Register(tracer);

                return tracer;
            });
```

**And below are the variables used to deploy image in both environments windows container and kubernetes -** 

`docker run --rm --link jaeger_local --env JAEGER_SERVICE_NAME="facility.default" --env JAEGER_AGENT_HOST=jaeger_local --env JAEGER_AGENT_PORT=6831 --env JAEGER_SAMPLER_TYPE=const --env JAEGER_SAMPLER_PARAM="1" --env JAEGER_TRACEID_128BIT=true --env JAEGER_REPORTER_LOG_SPANS=true -p 1561:80 facility`

Falco20019:
Fixes #143

Currently, B3 TraceId extraction only supports fallback to 64 bit. Since other tracers by now also support 128 bit, the behavior was aligned with the regular `TextMapCodec` again.

The `HexCodec` class was removed since it's redundant now with the `FromString` implementations of `SpanId` and `TraceId`. `B3TextMapCodec` earlier returned `null` on problems. Due to using the `ExceptionCatchingExtractorDecorator`, it's better to throw an exception that gets logged and still gracefully returns `null`.


yurishkuro:
don't know what github is smoking...

image
image

PR #150, opened 2019-07-11
Extending the range of supported versions of newtonsoft

SanSYS:
- Added support for projects which depends on old Newtonsoft Json library

- Now depended on Newtonsoft.Json 9.0.1 and above


SanSYS:
@Falco20019 @yurishkuro 
Hi! You'll help me very much if accept changes )

Falco20019:
@SanSYS I will release it as a patch version 0.3.3 since functionality is not changed. Should be on NuGET in the next 10 minutes.

PR #151, opened 2019-07-12
Invert dependency between Jaeger and Jaeger.Thrift NuGet packages 

### LINKED ISSUE #116: Invert dependency between Jaeger and Jaeger.Thrift NuGet packages

Falco20019:
- Fixes #116
- Related to #76 

This is a breaking change in a couple of aspects.

* Without setting-up `Configuration.SenderConfiguration.DefaultSenderResolver`, `RemoteReporter.Builder.Build()` will result in `NoopSender` instead of `UdpSender`
* Package structure changes, although the `Jaeger` NuGET package will resolve correctly
* This will result in the usage being compile-compatible but not assembly-compatible.

This solution is similar to how it was solved in Java.

## Things taken into account
I tried to avoid having to `Configuration.SenderConfiguration.DefaultSenderResolver` by using reflection. This works for .NET Framework 4.7 and .NET Core 2.1, but did not work with ASP.NET Core 2.1. The main problem is, that ASP.NET Core did not include the assembly in the AppDomain since it was not called directly from code. 

Another option would have been to use something like a `Jaeger.Initialize()` method that could be offered by the regular `Jaeger` package, but focusing on just a central default seemed like the most straight-forward and cleaner solution.


Falco20019:
@yurishkuro As #118 is solved now, I would like to finally trigger this change.

src/Jaeger.Core/Reporters/README.md
```diff
@@ -31,14 +31,19 @@ var reporter = new LoggingReporter(loggerFactory);
...
     .WithLoggerFactory(loggerFactory) // optional, defaults to no logging
     .WithMaxQueueSize(...)            // optional, defaults to 100
     .WithFlushInterval(...)           // optional, defaults to TimeSpan.FromSeconds(1)
-    .WithSender(...)                  // optional, defaults to UdpSender("localhost", 6831, 0)
+    .WithSenderResolver(...)          // optional, defaults to Configuration.SenderConfiguration.DefaultSenderResolver
+    .WithSender(...)                  // optional, defaults to SenderResolver.Resolve()
     .Build();
 ```
 
+For more information on sender resolution see the sender [README](src/Jaeger.Core/Senders/README.md)
```
yurishkuro, line 45: do you need the link to be `./src/...`?

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -239,7 +239,7 @@ public RemoteReporter Build()
                 }
                 if (_sender == null)
                 {
-                    _sender = new UdpSender();
+                    _sender = NoopSender.Instance;
```
yurishkuro, line 242: I would rather throw exception, null should not be valid, and Noop as default doesn't make much sense.

src/Jaeger.Core/Senders/README.md
```diff
@@ -0,0 +1,62 @@
+# Senders
```
yurishkuro, line 1: nit: personally, I prefer one large readme at the root, where you can search w/ Ctrl-F

src/Jaeger.Core/Senders/README.md
```diff
@@ -0,0 +1,62 @@
...
+If the factory to use is not found or was ambiguous due to no `SenderFactory` being defined, `NoopSender.Instance` is returned instead.
+
+If a factory was found, `ISenderFactory.GetSender()` will be used to get a configured instance of `ISender` to be returned to the caller of `SenderResolver.Resolve()`.
+
+```C#
+var senderResolver = new SenderResolver(loggerFactory)
+	.RegisterSenderFactory<ThriftSenderFactory>();
+var senderConfiguration = new Configuration.SenderConfiguration(loggerFactory)
+	.WithSenderResolver(senderResolver)	// optional, defaults to Configuration.SenderConfiguration.DefaultSenderResolver
+	.WithSenderFactory("thrift");		// optional if only one Factory registered to senderResolver
```
yurishkuro, line 24:
```suggestion
	.WithSenderFactory("thrift");		// optional when only one Factory registered to senderResolver
```

src/Jaeger.Core/Senders/SenderResolver.cs
```diff
@@ -0,0 +1,122 @@
...
+namespace Jaeger.Senders
+{
+    /// <summary>
+    /// Provides a way to resolve an appropriate <see cref="ISender"/>.
+    /// </summary>
+    public class SenderResolver
+    {
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly ILogger _logger;
+        private readonly IDictionary<string, ISenderFactory> _senderFactories;
```
yurishkuro, line 15: Trying to understand this part. In Java the "resolver" pattern does not require explicit registration, but auto-discovers the factories on the classpath. Is the intention the same here, or is manual registration always required? In which case, why support more than one registered factory?

src/Jaeger.Core/Reporters/RemoteReporter.cs
Falco20019: We also have an `NoopReporter` and a `NoopMetricsFactory`, so it would be natural to also include an `NoopSender`. But from a user-perspective, when a non-`NoopReporter` is chosen.

The only use case I can currently think of that this allows for is getting metrics without the need for a sender.

src/Jaeger.Core/Senders/README.md
Falco20019: We already had it this way with `Samplers` and `Reporters`. I usually try to avoid clutter and redundancy, but I can see your point. I will talk with some colleagues about their opinion and think about it :)

src/Jaeger.Core/Reporters/RemoteReporter.cs
Falco20019: BTW: Java is also using the NoopSender in this case. I will add a warning as Java is doing:
https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/internal/senders/SenderResolver.java#L60

src/Jaeger.Core/Senders/SenderResolver.cs
Falco20019: As written in the `Things taken into account` section of the first post, auto-discovery sadly not works with ASP.NET Core since it's not loading assemblies not referenced by the code (in comparison to .NET Core and .NET Framework). Therefore it's always required to keep it the same over all implementations.

Java also supports multiple senders here: https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/internal/senders/SenderResolver.java#L56

So I made sure to use the same JAEGER_SENDER_FACTORY environment variable. If only one is registered, it's not necessary to define it and just the one will be used. This should be the common use case. I liked the dictionary more than just an enumerable as mine is not generated by auto-discovery, as it helps with resolving the one set with JAEGER_SENDER_FACTORY.

src/Jaeger.Core/Reporters/README.md
Falco20019: Thanks for recognizing. Fixed the link.

src/Jaeger.Core/Senders/README.md
Falco20019: Thanks, updated. And I also replaced the constant string with `ThriftSenderFactory.Name` and added the note that `"thrift"` is to be used if `JAEGER_SENDER_FACTORY` is necessary.

src/Jaeger.Core/Senders/README.md
```diff
@@ -0,0 +1,62 @@
...
+var udpSender = new Configuration.SenderConfiguration(loggerFactory)
+	.WithAgentHost("jaeger-agent")
+	.WithAgentPort(6832)
+	.GetSender();
+
+// This creates a tracer, using an ISender generated through ThriftSenderFactory, depending on configuration
+var tracer = Configuration.FromEnv(loggerFactory).GetTracer();
+```
+
+# Notice
```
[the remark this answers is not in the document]
Falco20019: Added a note to the main README.

README.md
```diff
@@ -9,7 +9,15 @@
 This library is still under construction and needs to be peer reviewed as well as have features added.
 
 ## Usage
-This package contains everything you need to get up and running. If you want to report to a system such as Jaeger or Zipkin you will need to use their NuGet packages.
+This package contains everything you need to get up and running. The meta-package includes the following packages:
+* [Jaeger.Core](https://www.nuget.org/packages/Jaeger.Core/)
+* [Jaeger.Senders.Thrift](https://www.nuget.org/packages/Jaeger.Senders.Thrift/)
+
+### Senders
+The implementation of `Jaeger.Core` is agnostic to any reporting endpoint and could be even used without any in case of logfile reporting. For more information about the sender concept, have a look at the sender [README](src/Jaeger.Core/Senders/README.md).
```
yurishkuro, line 17: > src/Jaeger.Core/Senders/README.md

Are you sure these links work without the leading dot?  Maybe GH fixed it.

README.md
```diff
@@ -9,7 +9,15 @@
...
+
+This is a list of sender implementations known to the Jaeger team:
+* [Jaeger.Senders.Thrift](src/Senders/Jaeger.Senders.Thrift/README.md) (default, included in `Jaeger`)
```
yurishkuro, line 20: L23 - I find it a bit weird that the very first code example is not something that's actually useful. I would prefer to see "the right way" to create a real, prod-quality tracer.

yurishkuro:
you may need to squash all the commits to fix the DCO

Falco20019:
I squash and merge them using admin rights, so I don’t care about DCO upfront. I just make sure that I care about the warning and have it in the squashed :)

I will integrate your suggestions and merge it later today. Thanks again for the great input!

README.md
Falco20019: I also tried it yesterday with leading dot if in a sub folder but it’s not basing it correctly. In this case it‘s working since it‘s a relative path to the README and not relative to the document root.

Falco20019: Seems like this was introduced 2013: https://github.blog/2013-01-31-relative-links-in-markup-files/

README.md
Falco20019: It's useful to get the system up and running and making sure that your code is working without having dependencies to other things like environment, senders and so on. That's why I usually like to give simple stand-alone examples and increase the complexity on the way down.

This was also how I started to integrate Jaeger into our application and I found it annoying having to get a minimal example to work first :) Especially since there are lots of ways to get the configuration in on C#, not only environment variables.

src/Jaeger.Core/Senders/README.md
Falco20019: Our department wants to keep the documents separated as it allows us to reference them in context without having to link into only a chapter of a very long document.

PR #156, opened 2019-08-25
Ignore Jaeger internal traces in HttpSender (#154)

### LINKED ISSUE #154: Infinite span creation for Jaeger internal requests when OpenTracing.Contrib.NetCore and HttpSender are used
Hide Jaeger internal request traces in Jaeger UI to simplify trace searching for customers.

If JAEGER_ENDPOINT property is specified Jaeger uses HttpSender to send spans via Http directly to Jaeger Collector. [OpenTracing.Contrib.NetCore package](https://github.com/opentracing-contrib/csharp-netcore) uses DiagnosticSource to intercept each Http request and build a span. It leads to infinite span creation loop because each POST request to collector with span data creates another span and cause another POST request. Huge amount of such traces make it difficult to search real user traces.

изображение
изображение

Make OpenTracing.Contrib.NetCore ignore request to Jaeger Collector by specifing "ot-ignore" request property. 
It's already done in [DefaultHttpClient](https://github.com/jaegertracing/jaeger-client-csharp/blob/master/src/Jaeger/Util/DefaultHttpClient.cs) to hide receiving of sampling settings. The same could be done in [THttpClientTransport](https://github.com/jaegertracing/jaeger-client-csharp/blob/master/src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs)

See also [HttpHandlerDiagnosticOptions](https://github.com/opentracing-contrib/csharp-netcore/blob/master/src/OpenTracing.Contrib.NetCore/CoreFx/HttpHandlerDiagnosticOptions.cs) in OpenTracing.Contrib.NetCore

bocharovf:
Resolves #154 

Mark Jaeger client internal requests made by HttpSender with "ot-ignore" property which prevents requests from being instrumented by OpenTracing.Contrib.NetCore package.

### DIFF
--- a/src/Jaeger/Senders/HttpSender.cs
+++ b/src/Jaeger/Senders/HttpSender.cs
@@ -45,7 +45,14 @@ namespace Jaeger.Senders
                 customHeaders.Add("Authorization", builder.AuthenticationHeaderValue.ToString());
             }
 
-            _transport = new THttpClientTransport(collectorUri, customHeaders);
+            var customProperties = new Dictionary<string, object>
+            {
+                // Note: This ensures that internal requests from the tracer are not instrumented
+                // by https://github.com/opentracing-contrib/csharp-netcore
+                { "ot-ignore", true }
+            };
+
+            _transport = new THttpClientTransport(collectorUri, customHeaders, customProperties);
             _protocol = ProtocolFactory.GetProtocol(_transport);
         }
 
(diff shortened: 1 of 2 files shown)


Falco20019:
Thanks for fixing it. I currently have only little time.

Falco20019:
I just released it as 0.3.4 to NuGET, so that your problem gets fixed quickly.

PR #157, opened 2019-09-07
Default SamplingManager set to HttpSamplingManager

regeldso:
Default SamplingManager is set to HttpSamplingManager.

https://github.com/jaegertracing/jaeger-client-csharp/tree/master/src/Jaeger/Samplers#remote-controlled-sampler

Even default sampler is RemoteControlledSampler it doesn't have default SamplingManager as it mentioned in documentation that's why this situation produce an exeption during execution.

Added default SamplingManager, added one unit test.


Falco20019:
I just published it as v0.3.5 to NuGet. Should be available soon.

Issue #163, opened 2020-02-03
RemoteControlledSampler fails when Agent is on a different host from the application


hannahchan:
When the `JAEGER_AGENT_HOST` environment variable is set to a host other than the default localhost, the Jaeger.Samplers.RemoteControlledSampler fails with the following exception;

```
warn: Jaeger.Samplers.RemoteControlledSampler[0]
      Updating sampler failed
System.Net.Http.HttpRequestException: Cannot assign requested address
 ---> System.Net.Sockets.SocketException (99): Cannot assign requested address
   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean allowHttp2, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.CreateHttp11ConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.GetHttpConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
   at Jaeger.Util.DefaultHttpClient.MakeGetRequestAsync(String requestUri)
   at Jaeger.Samplers.HttpSamplingManager.GetSamplingStrategyAsync(String serviceName)
   at Jaeger.Samplers.RemoteControlledSampler.UpdateSampler()
```

I think it's trying to read the sampling configuration of the agent. When I change to the `const` sampler by setting the following environment variables, the agent starts receiving traces.

```
JAEGER_SAMPLER_TYPE: const
JAEGER_SAMPLER_PARAM: 1
```
This confirms that the network setup for the application and agent is correct and that the issue is in the RemoteControlledSampler class.

[...]

PR #165, opened 2020-02-14
Replace JAEGER_SAMPLER_MANAGER_HOST_PORT with JAEGER_SAMPLING_ENDPOINT

Falco20019:
- Implements https://github.com/jaegertracing/jaeger/issues/1849
- Fixes https://github.com/jaegertracing/jaeger-client-csharp/issues/163

- JAEGER_SAMPLER_MANAGER_HOST_PORT was changed to JAEGER_SAMPLING_ENDPOINT and contains a base uri now instead of host:port

### DIFF
--- a/src/Jaeger/Configuration.cs
+++ b/src/Jaeger/Configuration.cs
@@ -83,6 +83,11 @@ namespace Jaeger
         public const string JaegerSamplerParam = JaegerPrefix + "SAMPLER_PARAM";
 
         /// <summary>
+        /// The sampler manager host:port.
+        /// </summary>
+        public const string JaegerSamplerManagerHostPort = JaegerPrefix + "SAMPLER_MANAGER_HOST_PORT";
+
+        /// <summary>
         /// The url for the remote sampling conf when using sampler type remote.
         /// </summary>
         public const string JaegerSamplingEndpoint = JaegerPrefix + "SAMPLING_ENDPOINT";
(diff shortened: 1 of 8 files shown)


src/Jaeger/Configuration.cs
```diff
@@ -83,9 +83,9 @@ public class Configuration
         public const string JaegerSamplerParam = JaegerPrefix + "SAMPLER_PARAM";
 
         /// <summary>
-        /// The sampler manager host:port.
+        /// The url for the remote sampling conf when using sampler type remote.
         /// </summary>
-        public const string JaegerSamplerManagerHostPort = JaegerPrefix + "SAMPLER_MANAGER_HOST_PORT";
+        public const string JaegerSamplingEndpoint = JaegerPrefix + "SAMPLING_ENDPOINT";
```
yurishkuro: Usually we make these changes backward compatible by supporting the previous configuration for some period of time.

Falco20019: I only looked at the CPP implementation and saw that it was renamed. I had not seen the discussion there that it was not implemented before. I will add it on Monday and mark it deprecated.

yurishkuro: Ah, the reason it was a straight renaming in CPP is because the env var there was not used at all, so the functionality was never released and never available to users of the library.

Falco20019: I just readded the support and marked it as deprecated. If someone is using it, the logging will produce a warning at runtime.

src/Jaeger/Samplers/HttpSamplingManager.cs
```diff
@@ -31,15 +39,15 @@ internal SamplingStrategyResponse ParseJson(string json)
 
         public async Task<SamplingStrategyResponse> GetSamplingStrategyAsync(string serviceName)
         {
-            string url = "http://" + _hostPort + "/?service=" + Uri.EscapeDataString(serviceName);
```
yurishkuro: Interesting, this wasn't using the `/sampling` endpoint previously?  The results from `/` endpoint are using legacy Thrift-to-JSON conversion that used numbers for enums, instead of strings. Have you tested this change with an agent? I don't think crossdock tests cover the sampling endpoint.

Falco20019: I checked the code again. The enum is not checked at all, therefore it's not making problems. C# is deserializing the following data structure as property. So it's looking for the field names instead of the enum. But I will adjust the test cases to make sure this reflects the real situation.

PR #166, opened 2020-02-25
[Thrift] HttpClientHandler.ClientCertificates MissingMethodException

### LINKED ISSUE #105: [Thrift] HttpClientHandler.ClientCertificates not present in some .NET Framework versions

JnsSo:
This is an addition to the issue #105.
Seperated the call of handler.ClientCertificates.AddRange() to a single function.
Some APIs are not implemented in net461/net462, but they are available in
netStandard2.0, see: https://stackoverflow.com/a/46577035/10038915
Signed-off-by: Jonas S. j@tkaj.de

- Resolves #105 

- The additional "if-clause" does not solve the problem completely. Probably the just-in-time compiler tries to access the ClientCertificates property and so the exception gets thrown. I seperated the call of the not implemented property (eg. in .NET 4.6.1) to a single function to avoid the exception at JIT time.

### DIFF
--- a/src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
+++ b/src/Thrift/netcore/Thrift/Transports/Client/THttpClientTransport.cs
@@ -147,12 +147,23 @@ namespace Thrift.Transports.Client
             await _outputStream.WriteAsync(buffer, offset, length, cancellationToken);
         }
 
+        /// <summary>
+        /// Seperate function to avoid a MissingMethodException in .NET Framework 4.6.1
+        /// https://stackoverflow.com/a/46577035/10038915
+        /// Related to: https://github.com/jaegertracing/jaeger-client-csharp/issues/105
+        /// </summary>
+        /// <param name="handler">HttpClientHandler</param>
+        private void AddCertificates(HttpClientHandler handler)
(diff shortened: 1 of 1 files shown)


Falco20019:
I hope to release v0.3.7 soon. Only waiting for a last PR review on #165

PR #152, opened 2019-07-15
[WIP] Added crossdock integration tests

Falco20019:
## Which problem is this PR solving?
- Fixes #85 

## Short description of the changes
- Added crossdock support to jaeger-client-csharp
- Testing against csharp, go, java and python

Issue #85, opened 2018-06-04
Create crossdock integration tests


yurishkuro:
To address issues like https://github.com/jaegertracing/jaeger-client-csharp/issues/80#issuecomment-394416348

Falco20019:
It's a hard problem on Appveyor. Right now, we are using Windows Server 2016 build environment to build everything. Sadly, Appveyor does not support running Docker Containers for Linux on those machines. There seems to be an option if you buy nested virtualization but that's also a bit quirky...

They also work on a solution based on Windows Server 2019 and Docker EE which would allow using Windows + Linux containers. But as it looks now, the best option would be to just switch to an Ubuntu environment for our Builds since we are not really depending on using Windows right now AFAIK. But this will definetly take some more time.

PR #167, opened 2020-04-07
Added crossdock integration tests

Falco20019:
- Addresses #85 
- Follow-up #152

## Limitations
We sadly still can not create Linux images from AppVeyor, but at least we want to add the code and be locally able to test it.

- Added crossdock support to jaeger-client-csharp
- Testing against csharp, go, java and python


Falco20019:
@saschaholesch I would like to have a look with you together over the integration.

PR #169, opened 2020-04-14
Add GrpcSender for jaeger-collector

### LINKED ISSUE #139: Support for GrpcSender

Falco20019:
- Fixes #139
- Supersedes #141

- WIP until #167 is merged and code is checked

- The jaeger-collector got an gRPC endpoint. This can be used instead of `Thrift`. This add a separate package `Jaeger.Senders.Grpc` (like `Jaeger.Senders.Thrift`).

This can be used with `Jaeger.Core` to avoid having a dependency on `Thrift`. It includes no jaeger-agent sender and makes this also clear in the documentation. In comparison to #141, this also now supports `MaxPacketSize` like with `UdpSender`.


Falco20019:
Once #167 is merged, I will also add cross dock tests. For now, I only tested it manually against jaeger-collector in jaegertracing/all-in-one.

README.md
```diff
@@ -85,6 +86,8 @@ Property | Required | Description
 JAEGER_SERVICE_NAME | yes | The service name
 JAEGER_AGENT_HOST | no | The hostname for communicating with agent via UDP
 JAEGER_AGENT_PORT | no | The port for communicating with agent via UDP
+JAEGER_GRPC_TARGET | no | The target for communicating with collector via GRPC
+JAEGER_GRPC_ROOT_CERTIFICATE | no | The root certificate used to create a trusted TLS connection to the GRPC collector
```
yurishkuro, line 90: Could use some explanation - according to [docs][1], this is not using a client certificate pair, but only the CA certs.

[1]: https://grpc.github.io/grpc/csharp/api/Grpc.Core.SslCredentials.html#Grpc_Core_SslCredentials__ctor_System_String_

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
```diff
@@ -0,0 +1,153 @@
...
+
+namespace Jaeger.Senders.Grpc
+{
+    /// <summary>
+    /// GrpcSender provides an implementation to transport spans over HTTP using GRPC.
+    /// </summary>
+    public class GrpcSender : ISender
+    {
+        public const string DefaultCollectorGrpcTarget = "localhost:14250";
+        public const int MaxPacketSize = 65000;
```
yurishkuro, line 20: There's no expectation that gRPC messages are limited to 65000 bytes, this number only applies to UDP servers in the jaeger-agent. I am not sure this is sensible default, it's worth benchmarking. It could also be highly dependent on the network conditions, so probably worth exposing to the user via configuration.

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                maxPacketSize = MaxPacketSize;
+            }
+
+            _channel = new Channel(target, credentials);
+            _client = new CollectorService.CollectorServiceClient(_channel);
+            _maxSpanBytes = maxPacketSize;
+        }
+
+        public async Task<int> AppendAsync(Span span, CancellationToken cancellationToken)
```
yurishkuro, line 64: What does the use of async signify? Is this class supposed to provide thread safety for its internal state, like _spanBuffer?

src/Senders/Jaeger.Senders.Grpc/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Usage
+This package contains the `GrpcSenderFactory` for use with `SenderResolver`.
+
+To use this, include the following line before doing any other calls to the `Configuration` API.
+
+```C#
+Configuration.SenderConfiguration.DefaultSenderResolver = new SenderResolver(loggerFactory)
+	// You can add other ISenderFactory instances too
+	.RegisterSenderFactory<ThriftSenderFactory>();
```
yurishkuro, line 11:
```suggestion
	.RegisterSenderFactory<GrpcSenderFactory>();
```
?

src/Senders/Jaeger.Senders.Grpc/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+If you need to set `JAEGER_SENDER_FACTORY` (when multiple factories are registered), `GrpcSenderFactory.Name` corresponds to `"grpc"`.
+
+## Configuration
+Independent on what variables are set in `Configuration.SenderConfiguration`, a `GrpcSender` is created.
+
+```C#
+new SenderConfiguration(loggerFactory)
+	// Used with UdpSender:
```
yurishkuro, line 21: what does this mean?

src/Senders/Jaeger.Senders.Grpc/README.md
```diff
@@ -0,0 +1,39 @@
...
+	.WithGrpcRootCertificate(grpcRootCertificate); // Path to the root certificate
+```
+
+## Contained Senders
+This package contains one sender. `GrpcSender` for communicating with the `Jaeger Collector`.
+
+In containerized environments, you usually want to talk to the `Jaeger Agent` via UDP for fast communication with little overhead. This package does not allow communication with `Jaeger Agent`. So if you require that, use the `ThriftSenderFactory` instead.
+
+### GrpcSender
+Communicates with the `Jaeger Collector`. This sender offers authentication through client certificate using `GrpcRootCertificate`.
```
yurishkuro, line 32: I don't think it's accurate to refer to this as "client certificate" - root/CA cert only authenticates the signatory of the server cert. To support client cert, another pair must be provided in the config, e.g. https://grpc.github.io/grpc/csharp/api/Grpc.Core.SslCredentials.html#Grpc_Core_SslCredentials__ctor_System_String_Grpc_Core_KeyCertificatePair_

src/Senders/Jaeger.Senders.Grpc/protos/collector.proto
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2019 The Jaeger Authors.
```
yurishkuro, line 1: Are these straight copies from jaeger-idl, or modified somehow? If not modified, any reason not to use the git submodule as the other repos are doing? If modified, is it documented somewhere _how_ they are modified?

src/Jaeger.Thrift/Senders/Internal/ThriftUdpClientTransport.cs
```diff
@@ -9,7 +9,7 @@ namespace Jaeger.Thrift.Senders.Internal
 {
     public class ThriftUdpClientTransport : TTransport
     {
-        public const int MaxPacketSize = 65000; // TODO !!! Not yet used.
+        public const int MaxPacketSize = 65000;
```
yurishkuro, line 12: no other changes in this file, is it still unused?

Falco20019: Just found the left-over comment and saw that was wrong. It is in use as it will be given to ThriftSenderBase through UdpSender where it‘s used.

README.md
```diff
@@ -85,6 +86,8 @@ Property | Required | Description
 JAEGER_SERVICE_NAME | yes | The service name
 JAEGER_AGENT_HOST | no | The hostname for communicating with agent via UDP
 JAEGER_AGENT_PORT | no | The port for communicating with agent via UDP
+JAEGER_GRPC_TARGET | no | The target for communicating with collector via GRPC
```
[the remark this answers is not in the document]
Falco20019: In gRPC it‘s called „target“. This can be any valid target at only host, host:port, dns://host:port or since c-ares is supported dns://remote-dns/host:port. Or even Unix pipe. See https://github.com/grpc/grpc/blob/master/doc/naming.md

README.md
Falco20019: Thanks for pointing out. Misremembered it when documenting. This is the PEM content of the root certificate and not the path. Do you prefer the path or PEM content?

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
Falco20019: This has to be 4 MB (or a little less depending on process size) as that‘s the default for gRPC if not defined manually.

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
Falco20019: async only allows the usage of the await keyword in the method. It brings the compiler in a mode where it generates code returning a Task (promise) instead of the real value. Basically splitting up the method at each await into a asynchronous statemachine.

It has no meaning about safety :)

src/Senders/Jaeger.Senders.Grpc/README.md
Falco20019: Thanks, missed that on copy-pasting.

src/Senders/Jaeger.Senders.Grpc/README.md
Falco20019: Wanted to replace it with GrpcSender. Copy-Paste error.

src/Senders/Jaeger.Senders.Grpc/README.md
Falco20019: You are right, it’s only the PEM content of the root certificate used to authenticate the server certificate. Is Jaeger collector allowing client certificate usage / allows for setting a server certificate? If yes, we might want to add also a way to really set the client certificate.

src/Senders/Jaeger.Senders.Grpc/protos/collector.proto
Falco20019: Modified to remove the gogoproto stuff to allow Grpc.Tools to generate the cs files from proto. But I just saw that jaeger-idl is using docker and also generating csharp. Will use the generated ones then like with Thrift.

Falco20019: This is currently blocked by https://github.com/jaegertracing/jaeger-idl/issues/62

README.md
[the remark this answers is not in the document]
Falco20019: Added clarification. Hope it's fine for you.

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
Falco20019: Updated to 4 MB. Resolve if OK for you.

src/Senders/Jaeger.Senders.Grpc/protos/collector.proto
Falco20019: I locally generate them through jaeger-idl (from PR) and added the generated code to the class as `Jaeger.Communication.Grpc`.

Falco20019:
I left the conversations open for you to easier find the content. I resolved all but the `JAEGER_GRPC_ROOT_CERTIFICATE` topic. We need to decide if we want to support client certificates and if we want to have the content in the environment variable or the path. I assume the path would be better to use with `secrets`.

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
yurishkuro: What I mean was whether the use of async might cause a number of parallel executions of this method.

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
yurishkuro: 4MB sounds like a lot of data to buffer. Is there a timer-based flush in addition to size limit?

src/Senders/Jaeger.Senders.Grpc/README.md
yurishkuro: collector only has this for clients:
```
--collector.grpc.tls.client-ca string             Path to a TLS CA (Certification Authority) file used to verify certificates presented by clients
```

src/Senders/Jaeger.Senders.Grpc/protos/collector.proto
yurishkuro: does it mean what we can remove proto files from this PR?

README.md
yurishkuro: Don't have a preference. File sounds a bit simpler to manage, and that's what server options support.

README.md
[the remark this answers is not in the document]
yurishkuro: thx

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
Falco20019: Timer based flush is in RemoteReporter for all senders.

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
Falco20019: Ah yes, it‘s possible that the method is called in parallel, but since one sender is usual only used by one RemoteReporter, this should still be called in sequence and async then is only for interrupt handling.

src/Senders/Jaeger.Senders.Grpc/protos/collector.proto
Falco20019: Yes, they are removed in latest commits. Depending on if we are able to update jaeger-protobuf and jaeger-idl, I will either keep pre-generated code from idl in here or include it as submodule. Submodule would need us to also add more submodules to all referenced protos like gogo and the api annotations.

src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
yurishkuro: > and async then is only for interrupt handling.

That's what concerns me: RemoteReporter calls Append in one thread, but `await` defers the execution and might continue after the callback on another thread (but I may be misunderstanding the threading behavior in .NET with async).

src/Senders/Jaeger.Senders.Grpc/protos/collector.proto
Falco20019: This is still blocked by:
* https://github.com/jaegertracing/jaeger-idl/pull/63
* https://github.com/jaegertracing/docker-protobuf/pull/15

I don't see them to move fast. So I will use the pre-generated files for now to close this. Regenerating the protos will be painful until those are resolved.

README.md
Falco20019: Fixed it to use the path .This also allows us to use secrets in docker.

src/Senders/Jaeger.Senders.Grpc/README.md
Falco20019: Fixed. Thanks :)

Falco20019:
@yurishkuro I fixed all of the remaining topics. The client ca is also now using a string and should match the usage of `collector.grpc.tls.client-ca`. Thanks for all the input.

src/Communication/Jaeger.Communication.Grpc/Gogoproto/Gogo.cs
```diff
@@ -0,0 +1,291 @@
+// <auto-generated>
```
yurishkuro, line 1: Kinda sad that this is being generated into the library, I was hoping gogo annotations, since they only apply to Go generator, would be no-op elsewhere.

src/Jaeger.Core/Configuration.cs
```diff
@@ -57,6 +57,26 @@ public class Configuration
...
 
+        /// <summary>
+        /// The target used to locate the grpc collector endpoint.
+        /// </summary>
+        public const string JaegerGrpcTarget = JaegerPrefix + "GRPC_TARGET";
+
+        /// <summary>
+        /// The root certificate file used to check the server side certificate from GRPC collector (roots.pem).
+        /// </summary>
+        public const string JaegerGrpcRootCertificate = JaegerPrefix + "GRPC_ROOT_CERTIFICATE";
```
yurishkuro, line 68: Q: is it, strictly speaking, scoped to gRPC only? Wouldn't it be possible to use the same certificates with HTTPS?

src/Communication/Jaeger.Communication.Grpc/Gogoproto/Gogo.cs
Falco20019: Sadly not. And it's not even possible to generate it without manually removing them. That was the first approach... But hard to maintain.

src/Jaeger.Core/Configuration.cs
Falco20019: You are technically right. But right now, it's only used there. Do we have environment variables for HTTPS in other clients? If yes, I would try to keep it in sync.

yurishkuro: I don't believe mTLS is supported in any other clients.

Falco20019: Then I will for now just merge it. For 1.0.0 we might look into it again.

PR #179, opened 2020-06-15
Add Custom Random Number Generator

### LINKED ISSUE #177: Duplicate Span IDs because of default Random Number Generator

This library uses the default .NET Random() constructor, which initializes the seed with the current time. Since we are starting multiple processes in parallel at the same time, we have intermittently seen multiple processes sending duplicate span IDs to the server. This leads to bad data on the server. 

https://docs.microsoft.com/en-us/dotnet/api/system.random.-ctor?view=netcore-3.1#System_Random__ctor

Proposal 1) Use a combination of the current time and the current process ID to initialize Random()
Proposal 2) Use https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rngcryptoserviceprovider?view=netcore-3.1 to generate the initial seed for Random().

phxnsharp:

Resolves #177

Allow clients to pass in their own random number generator.

Note, I wasn't sure how you might want to handle package dependencies between Jaeger and Jaeger.Util, so to be safe I duplicated the delegate definition to avoid either 1) Exposing Jaeger.Utils to the client or 2) Creating a circular dependency between Jaeger and Jaeger.Utils. If you don't care about those, one of the definitions can be removed.

Signed-off-by: phxnsharp <nsharp@phoenix-int.com>

### DIFF
--- a/src/Jaeger.Core/SpanBuilder.cs
+++ b/src/Jaeger.Core/SpanBuilder.cs
@@ -164,7 +164,7 @@ namespace Jaeger
 
         private SpanContext CreateNewContext(string debugId)
         {
-            TraceId traceId = TraceId.NewUniqueId(_tracer.UseTraceId128Bit);
+            TraceId traceId = TraceId.NewUniqueId(_tracer);
             SpanId spanId = new SpanId(traceId);
 
             var flags = SpanContextFlags.None;
@@ -247,7 +247,7 @@ namespace Jaeger
 
             return new SpanContext(
                 preferredReference.TraceId,
-                SpanId.NewUniqueId(),
+                SpanId.NewUniqueId(_tracer),
                 preferredReference.SpanId,
                 // should we do OR across passed references?
                 preferredReference.Flags,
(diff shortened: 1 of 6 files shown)


src/Jaeger.Core/SpanBuilder.cs
```diff
@@ -247,7 +247,7 @@ private SpanContext CreateChildContext()
 
             return new SpanContext(
                 preferredReference.TraceId,
-                SpanId.NewUniqueId(),
+                SpanId.NewUniqueId(_tracer),
```
yurishkuro, line 250: this may result in 128bit span ID, which are not supported.

phxnsharp: While both SpanId and TraceId now take a Tracer as context, only TraceId has the code to optionally create 128 bit spans. This pre-existing test proves that:

https://github.com/jaegertracing/jaeger-client-csharp/blob/5fe4c9becb4653262ed41ecc151b264533852ca1/test/Jaeger.Core.Tests/SpanTests.cs#L54

phxnsharp:
Thanks! #180 gives us what we need.

PR #180, opened 2020-06-15
Initialize Random Harder

### LINKED ISSUE #177: Duplicate Span IDs because of default Random Number Generator

phxnsharp:

Resolves #177

Fix for duplicate IDs from multiple processes started at the same time. Use a cryptographic
random number generator instead of current time to seed Random.

This can be used instead of or as well as #179, they do not conflict. This commit makes the default random number generation safe against multiple processes launching at the same time. I'm sure for most of your users this is quite unlikely as their processes tend to be long lived, and don't start very often. However, if it ever does happen it will be fatal. This random number generator is used for nothing else, so all trace and span IDs created by both processes will collide for the life of the processes.

Signed-off-by: phxnsharp <nsharp@phoenix-int.com>

### DIFF
--- a/src/Jaeger.Core/Util/Utils.cs
+++ b/src/Jaeger.Core/Util/Utils.cs
@@ -1,11 +1,24 @@
 using System;
 using System.Net;
+using System.Security.Cryptography;
 
 namespace Jaeger.Util
 {
     public static class Utils
     {
-        private static readonly Random Random = new Random();
+        private static readonly Random Random;
+
+        static Utils()
+        {
+            // Really initialize the random number generator so that multiple processes
+            // starting at the same time do not duplicate IDs.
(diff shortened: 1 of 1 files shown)


Falco20019:
Thanks for your PR. I'm currently busy but hope to have a look the next days!

PR #187, opened 2020-07-20
Update ISSUE_TEMPLATE.md

Falco20019:
- We get lots of tickets related to `OpenTracing.Contrib.NetCore` issues

- Add a note to check if this really belongs here or to `OpenTracing.Contrib.NetCore`.


yurishkuro:
@Falco20019 consider if making specific templates ([example](https://github.com/jaegertracing/jaeger/issues/new/choose)) makes sense, e.g. there could be an explicit link to the other repo.

PR #191, opened 2020-09-08
Fix blocking code

### LINKED ISSUE #189: RemoteReporter.cs blocking code issue
Hello guys,

First of all, thank you for all the work on this library. We are using Jaeger and it's a life saver. Recently, we had some issues with performance and during the analyses we found that jaeger C# is blocking our thread.

 In our case, we have a .net core application inside a docker container, on a k8s cluster. When we put a limit on the container to be exactly 1 CPU, we see some performance issues. After several investigations, one of the things we found out, is that we have thread blocking on jaeger code.

JaegerBlockingTrace

After looking at the source code, I think this could easily be avoided. My proposal is to change the method ProcessQueueLoop()
in RemoteReporter.cs to be async, so that when command.ExecuteAsync() is called it can be awaited and not to use Wait which is blocking.

Also, the StartNew code could be replaced with 

`_queueProcessorTask = Task.Run(async () =>
            {
                await ProcessQueueLoop();
            });`

BojanPantovic1989:
-Fixes #189

- replace blocking collection with BufferBlock to make everything async

### DIFF
--- a/src/Jaeger.Core/Reporters/RemoteReporter.cs
+++ b/src/Jaeger.Core/Reporters/RemoteReporter.cs
@@ -46,17 +46,9 @@ namespace Jaeger.Reporters
 
         public void Report(Span span)
         {
-            bool added = false;
-            try
-            {
-                // It's better to drop spans, than to block here
-                added = _commandQueue.Post(new AppendCommand(this, span));
-            }
-            catch (InvalidOperationException)
-            {
-                // The queue has been marked as IsAddingCompleted -> no-op.
-            }
-
+            // It's better to drop spans, than to block here
+            var added = _commandQueue.Post(new AppendCommand(this, span));
+            
             if (!added)
             {
                 _metrics.ReporterDropped.Inc(1);
--- a/src/Jaeger.Core/Reporters/RemoteReporter.cs
+++ b/src/Jaeger.Core/Reporters/RemoteReporter.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
 using Jaeger.Exceptions;
 using Jaeger.Metrics;
 using Jaeger.Senders;
@@ -18,7 +19,7 @@ namespace Jaeger.Reporters
         public const int DefaultMaxQueueSize = 100;
         public static readonly TimeSpan DefaultFlushInterval = TimeSpan.FromSeconds(1);
 
-        private readonly BlockingCollection<ICommand> _commandQueue;
+        private readonly BufferBlock<ICommand> _commandQueue;
         private readonly Task _queueProcessorTask;
         private readonly TimeSpan _flushInterval;
         private readonly Task _flushTask;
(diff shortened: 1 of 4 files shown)


BojanPantovic1989:
@Falco20019 I wasn't able to test using 0.4.1, but I applied changes to make everything async.

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -47,7 +50,7 @@ public void Report(Span span)
             try
             {
                 // It's better to drop spans, than to block here
-                added = _commandQueue.TryAdd(new AppendCommand(this, span));
+                added = _commandQueue.Post(new AppendCommand(this, span));
             }
             catch (InvalidOperationException)
```
phxnsharp, line 55: According to docs, .Post doesn't throw InvalidOperationException so this may now be not needed?

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -64,7 +67,7 @@ public async Task CloseAsync(CancellationToken cancellationToken)
         {
             // Note: Java creates a CloseCommand but we have CompleteAdding() in C# so we don't need the command.
```
phxnsharp, line 68: Comment refers to old CompleteAdding() instead of the new Complete() method

phxnsharp:
Looks great to me! Looks like maybe something went wrong the continuous integration above but I had written test cases for the original fix. If those still pass, I'm happy. 

Thanks!

src/Jaeger.Core/Reporters/RemoteReporter.cs
BojanPantovic1989: removed

src/Jaeger.Core/Reporters/RemoteReporter.cs
BojanPantovic1989: @phxnsharp Maybe I should use SendAsync instead of Post?

phxnsharp: By my reading, we want Post since if the queue is full, we want it to drop the command. My comment is just that InvalidOperationException is no longer thrown, so the catch clause may be dead code.

BojanPantovic1989: Yeah, I removed the dead code, I was just thinking because with SendAsync we postpone the adding to the queue if it is full.

phxnsharp: I'm not sure that makes sense, given the purpose of the code though. The idea is that it is meant to be low profile on your instrumented code. If we have failed to keep up and the queue is full, it is intentional that the packets get lost. Otherwise you'll just start filling up stack traces and Tasks instead of the queue I think.

Falco20019:
The CI failed because there were some deprecation warnings (treated as errors) from NuGET. I just fixed them. I will rerun the CI.

Falco20019 (approved):
LGTM.

@phxnsharp Thanks for doing the review on this one! Tests are green now after fixing the pipeline.

BojanPantovic1989:
Hi @Falco20019 I tried fixing it but I failed, and I am pretty sure I made things worse.

Falco20019:
I'm testing it a bit more right now. I found that current `0.4.1` has problems if something on the sender goes wrong. Since we use `command.ExecuteAsync().Wait();`, this will throw an `AggregateException` instead of `SenderException`. This PR should fix it since it's using async/await again.

EngRajabi:
Do not publish the version ??
We have a problem with version 0.4.1

Falco20019:
@EngRajabi Sorry, I missed that this is still open. I will release it now.

EngRajabi:
@Falco20019 thanks

PR #194, opened 2020-11-08
async flush

### LINKED ISSUE #193: buffer block size
## Problem - If the maximum number of buffer blocks is filled. You can no longer run the update command
If the maximum number of complete blocks is filled. You can no longer run the update command.
It is possible that in some cases we reach the maximum number of block items and can no longer order the update. This situation may occur when, for example, the server is not available

EngRajabi:
# Which problem is this PR solving?
- fix #193 (It is possible that in some cases we reach the maximum number of block items and can no longer order the update. This situation may occur when, for example, the server is not available)
- add async flush(Add async flush that performs operations optimally)

- add system.timer.timer for async flush with interval

### DIFF
--- a/src/Jaeger.Core/Reporters/RemoteReporter.cs
+++ b/src/Jaeger.Core/Reporters/RemoteReporter.cs
@@ -70,7 +70,7 @@ namespace Jaeger.Reporters
 
             if (_commandQueue.Count >= _maxQueueSize)
             {
-                var t = FlushAsync();
+                var t = FlushAsync(new CancellationTokenSource(10000).Token);
             }
 
         }
@@ -120,9 +120,9 @@ namespace Jaeger.Reporters
             }
         }
 
-        internal async Task FlushAsync()
+        internal async Task FlushAsync(CancellationToken token)
         {
-            await Task.Run(Flush);
+            await Task.Run(Flush, token);
         }
 
         internal void Flush()
--- a/src/Jaeger.Core/Reporters/RemoteReporter.cs
+++ b/src/Jaeger.Core/Reporters/RemoteReporter.cs
@@ -3,6 +3,7 @@ using System.Collections.Concurrent;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Dataflow;
+using System.Timers;
 using Jaeger.Exceptions;
 using Jaeger.Metrics;
 using Jaeger.Senders;
@@ -22,10 +23,11 @@ namespace Jaeger.Reporters
         private readonly BufferBlock<ICommand> _commandQueue;
         private readonly Task _queueProcessorTask;
         private readonly TimeSpan _flushInterval;
-        private readonly Task _flushTask;
+        private readonly int _maxQueueSize;
         private readonly ISender _sender;
         private readonly IMetrics _metrics;
         private readonly ILogger _logger;
+        private System.Timers.Timer _timer;
 
         internal RemoteReporter(ISender sender, TimeSpan flushInterval, int maxQueueSize,
             IMetrics metrics, ILoggerFactory loggerFactory)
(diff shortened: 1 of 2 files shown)


EngRajabi:
@Falco20019

Falco20019:
@EngRajabi Will try to have a look today or tomorrow. I saw it yesterday but didn’t have time.

vahidsing008:
I reviewed changes and there is no problem.

Falco20019:
I wanted to avoid using a separate timer and use a cancellation token instead to stop the running work. This way, the queue will not run over (only when it should). Otherwise it will create lots of pending tasks over time too.

EngRajabi:
My changes pursued two goals
First, if the block value reaches a maximum, it is no longer possible to send the update command.
My second goal was to add a command separate from sending information to the server.
This means that we send information to the server in two modes.
1- Reaching the number
2. Reaching the time
In either case, we must send the information to the server. In either case, the jaeger center server may be involved or have a network problem, and this should not affect the microservice process.
I will add in the codes from cancellationToken with a timeout of 10 seconds soon

EngRajabi:
Add cancellation token for thread

Falco20019:
@yurishkuro How is the go client handling this? I checked it, and it also seems to have a cyclic queue triggering the sending. Is there a timeout in the transports or is it only starting them in a cycle and let them run indefinitely?

@EngRajabi I'm consulting with Yuri to make sure our solution feels similar to the other clients. I see that the go client is also using a timer internally. I just wasn't able to quickly find out if or how pending transports are cancelled.

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -22,37 +23,56 @@ public class RemoteReporter : IReporter
...
+
+        private void InitializeFlushTimer()
+        {
+            _timer = new System.Timers.Timer
+            {
+                AutoReset = false,
+                Interval = _flushInterval.TotalMilliseconds
+            };
+            _timer.Elapsed += FlushOnTimer;
+            _timer.Start();
```
Falco20019, line 58: The timer is not stopped in `CloseAsync`.

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -88,25 +108,34 @@ public async Task CloseAsync(CancellationToken cancellationToken)
...
+            }
+            finally
+            {
+                _timer.Start();
+            }
+        }
+
+        internal async Task FlushAsync(CancellationToken token)
+        {
+            await Task.Run(Flush, token);
```
Falco20019, line 125: You should never use `async/await` on a simple `Task`, as this is creating a two state state-machine which is only unwrapping and rewrapping the result. Just return the `Task` and remove the `async` and `await` keywords.

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -22,37 +23,56 @@ public class RemoteReporter : IReporter
...
             // It's better to drop spans, than to block here
             var added = _commandQueue.Post(new AppendCommand(this, span));
-            
+
             if (!added)
             {
                 _metrics.ReporterDropped.Inc(1);
             }
+
+            if (_commandQueue.Count >= _maxQueueSize)
```
Falco20019, line 71: This would defy the purpose of the queue size. We ONLY want to flush on the time limit. The queue size is to ensure there are NEVER more elements around. This includes the ones being flushed.

src/Jaeger.Core/Reporters/RemoteReporter.cs
```diff
@@ -88,25 +108,34 @@ public async Task CloseAsync(CancellationToken cancellationToken)
...
-        private async Task FlushLoop()
-        {
-            // First flush should happen later so we start with the delay
-            do
-            {
-                await Task.Delay(_flushInterval).ConfigureAwait(false);
-                Flush();
-            }
-            while (!_commandQueue.Completion.IsCompleted);
+            var flushCommand = new FlushCommand(this);
```
Falco20019, line 135: This must only add the command to the processing queue as only the processor should flush.

Falco20019 (changes requested):

Falco20019:
I assume the main problem is not the `RemoteReporter`, as this is implemented in accordance to the go client and behaves as specified and expected. I only assume that there should be a timeout on the `Sender` which cancels it's flushing.

src/Jaeger.Core/Reporters/RemoteReporter.cs
EngRajabi: If the server is down and we have reached the maximum number of queues. How is our update command executed?

EngRajabi (commented):
resolve review.
But I have a question in one case
If the server is down and we have reached the maximum number of queues. How is our update command executed?

src/Jaeger.Core/Reporters/RemoteReporter.cs
Falco20019: It will just skip the flush and add it in the next `FlushLoop` cycle. Since the `FlushCommand` is blocking the `ProcessQueueLoop` until it returns, once the `Sender.Flush` is timing out properly, that `FlushCommand` will return properly and the queue is emptied with all remaining `AppendCommand` and `FlushCommand` being executed right afterwards.

Falco20019:
I assume you mean the `AppendCommand` with "update", right?

EngRajabi:
No, I mean this. FlushCommand

Falco20019:
This flush command should be skipped if the queue is empty, as the `FlushLoop` will add another one whenever there is space for it on the next cycle, just as the comment explains:
https://github.com/jaegertracing/jaeger-client-csharp/blob/master/src/Jaeger.Core/Reporters/RemoteReporter.cs#L96-L98

yurishkuro:
@Falco20019 not sure if I fully follow your question. The basic pattern in the clients is this:
  * Finished spans added to an internal bounded queue. If the queue is full spans are dropped.
  * A background thread reads off the queue and assembles spans in a batch. The batch is flushed to the backend when it's either large enough, of a certain timeout has expired (to avoid holding spans for a long time if their rate of arrival is very low).

Originally this was all designed for UDP, so a timeout was not an issue. With the addition of HTTP transports the implementations may vary. I'd say ideally the HTTP submission should be async from the background thread itself, as long as the batch buffer is properly freed for the next batch while the submission is ongoing.

Also, there is some variety in how the timeouts are handled, some clients use command pattern in the queue.

PR #210, opened 2021-04-19
Limit UDPs max span bytes when using Thrift

### LINKED ISSUE #199: Thrift.UdpSender not sending Span even though it's under MaxSpanBytes
I am attempting to send Spans to the Jaeger Agent via the Thrift.UdpSender that are over 65000 bytes.

I have my Jaeger Agent configured to accept Spans that are quite large via the following Agent CLI parameter:
```
--processor.jaeger-compact.server-max-packet-size=1048576
```

I then create a Span in my ASP.NET Core code and call `_tracer.ActiveSpan?.Log()` to log some fairly large data to the Span. This results in a Span that is over the default 65000 bytes limit for the `Thrift.UdpSender` so when I instantiate my UdpSender, I pass in a higher value for the `maxPacketSize` parameter (e.g. 1048576 as shown here):
```csharp
	var sender = new Jaeger.Senders.Thrift.UdpSender("localhost", 6831, 1048576);

	var reporter = new RemoteReporter.Builder().WithLoggerFactory(loggerFactory).WithSender(sender).Build();
	var initialSampler = new ConstSampler(true);
	var sampler = new RemoteControlledSampler.Builder(serviceName)
		.WithLoggerFactory(loggerFactory)      // optional, defaults to no logging
		.WithInitialSampler(initialSampler)    // optional, defaults to ProbabilisticSampler(0.001D)
		.WithPollingInterval(pollingInterval)  // optional, defaults to TimeSpan.FromMinutes(1)
		.WithSamplingManager(samplingManager)  // optional, defaults to HttpSamplingManager("localhost:5778")
		.Build();
```

Unfortunately, even though my size limit is set to well over 65000 characters and I can see `sender.MaxSpanBytes = 1048543`, my Span is still not sent to my Agent. I can tell this because I have my Agent running with debug level output and if I decrease the size of the Span enough, I see those being sent in the Agent logs (but not the larger ones) -- for example:
```
jaeger-agent_1  | {"level":"debug","ts":1608831249.214059,"caller":"processors/thrift_processor.go:120","msg":"Span(s) received by the agent","bytes-received":63662}
jaeger-agent_1  | {"level":"debug","ts":1608831254.2172835,"caller":"processors/thrift_processor.go:120","msg":"Span(s) received by the agent","bytes-received":64677}
```

So it seems even though I am specifying a larger packet size limit, it is still not being respected. I also have the log level set for my .NET Core application to log Trace level output for `Jaeger` and `Jaeger.Senders` namespaces but I see no log output to the console for these Spans:
```json
  "Logging": {
    "LogLevel": {
      "Default": "Information",
      "OpenTracing": "Debug",
      "Jaeger": "Trace",
      "Jaeger.Senders": "Trace",
      "Microsoft.AspNetCore": "Warning",
      "Microsoft.Hosting": "Warning"
    }
  },
```

Strangely, I tried setting the max size limit to a much lower value to see if I could get the ThriftSender to log this exception but I still am not seeing that:
```csharp
throw new SenderException($"ThriftSender received a span that was too large, size = {spanSize}, max = {MaxSpanBytes}", null, 1);
```

So I can't tell where the Span is actually being dropped. Am I doing something wrong with my sender setup / DI injection here that would reveal these logs and help me troubleshoot myself or is something else going wrong like the `maxPacketSize` not being respected? I looked through the jaeger-client-csharp code and I am not seeing an obvious issue.

Respect the `maxPacketSize` passed in to UdpSender.

Falco20019:
- Fixes #199

- Values above 64k can lead to lost traces as shown by #199. To avoid users running into this issue when manually instantiating the `UdpSender`, the value is now limited.

### DIFF
--- a/src/Communication/Jaeger.Communication.Thrift/Senders/ThriftSenderBase.cs
+++ b/src/Communication/Jaeger.Communication.Thrift/Senders/ThriftSenderBase.cs
@@ -21,8 +21,8 @@ namespace Jaeger.Thrift.Senders
 
         protected int MaxSpanBytes { get; }
 
-        /// <param name="protocolType">Protocol type (compact or binary)</param<
-        /// <param name="maxPacketSize">If 0 it will use default value <see cref="ThriftUdpTransport.MAX_PACKET_SIZE"/>.</param>
+        /// <param name="protocolType">Protocol type (compact or binary)</param>
+        /// <param name="maxPacketSize">If 0 it will use default value <see cref="ThriftUdpClientTransport.MaxPacketSize"/>.</param>
         public ThriftSenderBase(ProtocolType protocolType, int maxPacketSize)
         {
             switch (protocolType)
--- a/src/Senders/Jaeger.Senders.Thrift/UdpSender.cs
+++ b/src/Senders/Jaeger.Senders.Thrift/UdpSender.cs
@@ -36,11 +36,10 @@ namespace Jaeger.Senders.Thrift
 
         /// <param name="host">If empty it will use <see cref="DefaultAgentUdpHost"/>.</param>
         /// <param name="port">If 0 it will use <see cref="DefaultAgentUdpCompactPort"/>.</param>
-        /// <param name="maxPacketSize">If 0 it will use <see cref="ThriftUdpClientTransport.MaxPacketSize"/>.</param>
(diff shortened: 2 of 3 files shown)


Falco20019:
Once this is merged, I will finally release `1.0.0`.

yurishkuro (approved):

PR #213, opened 2021-05-27
Fix issues with Jaeger running on .NET Framework

### LINKED ISSUE #211: TypeLoadException when using latest version of Apache Thrift (0.14.1)

Using distributed tracing with Jaeger and UDP sender in .NET 5.0 application

A minimal console app written in C# works fine as long as ApacheThrift version 0.13.x or earlier is used. Recently Thrift was updated, and if I reference version 0.14.x (I tested specifically 0.14.1), then the application crashes with the following exception:

> System.TypeLoadException: Method 'get_Configuration' in type 'Jaeger.Thrift.Senders.Internal.ThriftUdpClientTransport' from assembly 'Jaeger.Communication.Thrift, Version=1.0.0.0, Culture=neutral, PublicKeyToken=9516061555a213a8' does not have an implementation.
>    at Jaeger.Senders.Thrift.UdpSender..ctor(String host, Int32 port, Int32 maxPacketSize)
> 

The sample application code:

```
using Jaeger;
using Jaeger.Senders.Thrift;
using Microsoft.Extensions.Logging;
namespace JaegerClient
{
    class Program
    {
        static void Main(string[] args)
        {
            var loggerFactory = new LoggerFactory();
            var senderConfiguration = new Configuration.SenderConfiguration(loggerFactory)
                .WithSender(new UdpSender());
            var reporterConfiguration = new Configuration.ReporterConfiguration(loggerFactory)
                .WithSender(senderConfiguration);
            var tracer = (Tracer)new Configuration("new-service", loggerFactory)
                .WithReporter(reporterConfiguration)
                .GetTracer();
            using (var scope = tracer.BuildSpan("foo").StartActive(true))
            {
                System.Threading.Thread.Sleep(1000);
            }
            tracer.Dispose();
        }
    }
}

```

Downgrading ApacheThrift library to 0.13.0.1 fixes the problem.

### LINKED ISSUE #212: Could not load assembly Thrift under .net framework 4
**Describe the bug**
When a project targeting framework net472 initializes Jaeger.Senders.Thrift.UdpSender it gets exception "System.IO.FileLoadException: Could not load file or assembly 'Thrift, Version=0.13.0.0, Culture=neutral, PublicKeyToken=5792536a45833b9f' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)"

**To Reproduce**
Please check example project: https://github.com/rechkalov/JaegerDemo/
It works with TargetFramework "netcoreapp3.1" or "net5.0", but not with "net472" or "net48".

**Version (please complete the following information):**
 - Jaeger versions: 1.0.0 to 0.4.0

**Additional context**
Looks like that the problem is with ApacheThrift package version 0.13.0.1. It contains assembly for .net 4.5 with name Thrift45.dll (not Thrift.dll like versions for net35 or netstandart2.0). Probably updating to next version will fix the issue.
It can not be solved by referencing newer ApacheThrift version on client side because of they changed some signatures (#211)

Falco20019:
- Fixes #211
- Fixes #212

- Updated `ApacheThrift` to 0.14
- Updated `Microsoft.Extensions.Logging.Abstractions` to 3.1
- Resolves dependency resolution between .NET Standard and .NET Framework


examples/Jaeger.Example.WebApi.NetFx/App_Start/AutofacConfig.cs
```diff
@@ -13,6 +16,10 @@ public class AutofacConfig
 
         static AutofacConfig()
         {
+            // This is necessary to pick the correct sender, otherwise a NoopSender is used!
```
yurishkuro, line 19: doesn't seem related to the PR directly, is this just an oversight from the previous resolver refactoring?

examples/Jaeger.Example.WebApi/Startup.cs
```diff
@@ -30,10 +33,16 @@ public void ConfigureServices(IServiceCollection services)
             // Adds the Jaeger Tracer.
             services.AddSingleton<ITracer>(serviceProvider =>
             {
-                string serviceName = serviceProvider.GetRequiredService<IWebHostEnvironment>().ApplicationName;
+                var serviceName = serviceProvider.GetRequiredService<IWebHostEnvironment>().ApplicationName;
+                var loggerFactory = serviceProvider.GetRequiredService<ILoggerFactory>();
```
yurishkuro, line 37: just curious, if .net already has this mechanism of service providers, why did we need a custom SenderConfiguration.DefaultSenderResolver? Couldn't we register ThriftSenderFactory with service provider?

yurishkuro (approved):

examples/Jaeger.Example.WebApi.NetFx/App_Start/AutofacConfig.cs
Falco20019: Yep, oversight from when we split up the packages. Since most people either use the NetCore package for ASP.NET (which has it‘s own example over there) or directly the configure options as in the main readme, nobody recognized.

examples/Jaeger.Example.WebApi/Startup.cs
Falco20019: That‘s only for ASP.NET (website stuff), but not for Desktop or Console applications. There are some dependency injection frameworks available, but I didn‘t want to force users to use one, since there is no gold standard.

PR #221, opened 2021-06-15
Update versions in example

### LINKED ISSUE #220: The type or namespace name 'Thrift' does not exist in the namespace 'Jaeger.Senders'
**Describe the bug**
`Jaeger.Example.WebApi/Startup.cs(4,22): error CS0234: The type or namespace name 'Thrift' does not exist in the namespace 'Jaeger.Senders' (are you missing an assembly reference?)`

**To Reproduce**
1. clone repo locally
2. navigate to examples/Jaeger.Example.WebApi
3. run `dotnet run`
4. receive error `CSC : error CS8102: Public signing was specified and requires a public key, but no public key was specified. [aeger-client-csharp/examples/Jaeger.Example.WebApi/Jaeger.Example.WebApi.csproj]`
5. replace relative path package reference with Nuget package reference as described in the comment on lines 10-14 in Jaeger.Example.WebApi.csproj
6. receive error `Jaeger.Example.WebApi/Startup.cs(4,22): error CS0234: The type or namespace name 'Thrift' does not exist in the namespace 'Jaeger.Senders' (are you missing an assembly reference?)`

**Expected behavior**
The app runs.

**Screenshots**

**Version (please complete the following information):**
 - OS: macOS
 - Jaeger version: [e.g. 1.8]
 - Deployment: [e.g. bare metal, Docker, Kubernetes, ...]

**What troubleshooting steps did you try?**
Try to follow https://www.jaegertracing.io/docs/latest/troubleshooting/ and describe how far you were able to progress and/or which steps did not work.

**Additional context**

Falco20019:
- Resolves #220 

- The example had an old version. When following the instructions for standalone (using the NuGet package instead of the `ProjectReference`), it used a version not matching the code anymore.

### DIFF
--- a/examples/Jaeger.Example.WebApi/Jaeger.Example.WebApi.csproj
+++ b/examples/Jaeger.Example.WebApi/Jaeger.Example.WebApi.csproj
@@ -5,13 +5,13 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="OpenTracing.Contrib.NetCore" Version="0.6.2" />
+    <PackageReference Include="OpenTracing.Contrib.NetCore" Version="0.7.1" />
 
     <!-- NOTE:
         In case you're building this application standalone, use below PackageReference 
         with an appropriate version instead of relative-path ProjectReference 
         to Jaeger project -->
-    <!-- PackageReference Include="Jaeger" Version="0.2.1" /-->
+    <!-- PackageReference Include="Jaeger" Version="1.0.2" /-->
   </ItemGroup>
 
   <ItemGroup>


yurishkuro (approved):
Is it possible to have a CI step that at least compiles the example?

Falco20019:
It should be compiled and should have worked. Only when the NuGet package is used instead of the local reference, the problem would occured. I was not able to reproduce the other issue that #220 mentioned.

examples/Jaeger.Example.WebApi/Jaeger.Example.WebApi.csproj
```diff
@@ -5,13 +5,13 @@
...
   <ItemGroup>
-    <PackageReference Include="OpenTracing.Contrib.NetCore" Version="0.6.2" />
+    <PackageReference Include="OpenTracing.Contrib.NetCore" Version="0.7.1" />
 
     <!-- NOTE:
         In case you're building this application standalone, use below PackageReference 
         with an appropriate version instead of relative-path ProjectReference 
         to Jaeger project -->
-    <!-- PackageReference Include="Jaeger" Version="0.2.1" /-->
+    <!-- PackageReference Include="Jaeger" Version="1.0.2" /-->
```
Falco20019, line 14: This line is commented out in favor of the `ProjectReference` further down (which is used in CI).

Issue #225, opened 2021-08-11
gRPC Sender has maxPacketSize parameter but it cannot exceed gRPC core client default max message length


js8080:
**Describe the bug**
The gRPC Sender in `Jaeger.Senders.Grpc.GrpcSender` has a constructor that takes the parameter `int maxPacketSize` and subsequent code in the GrpcSender ensures the Span being reported does not exceed this size. However, [the core gRPC Client being used](https://github.com/grpc/grpc) has its own default max message length of 4mb (`4*1024*1024`). This default max message length can be overridden by setting the `grpc.max_send_message_length` ChannelOption but unfortunately, the GrpcSender does not do this today. Therefore, if you try to use a `maxPacketSize` value greater than 4,194,304, your Span will not be reported to the Collector.

**To Reproduce**
1. When constructing your GrpcSender (such as in your `Startup.ConfigureServices()` method), pass a value larger than 4mb (4,194,304 bytes) via the `maxPacketSize` parameter.
2. Attempt to report a Span that is larger than this size.
3. The Span will not be reported to the gRPC Collector.

**Expected behavior**
The gRPC Core client allows overriding the `grpc.max_send_message_length` setting so we should set this ChannelOption when constructing the gRPC Channel.

**Screenshots**
n/a

**Version (please complete the following information):**
 - OS: Linux
 - Jaeger version: Jaeger v1.21.0 but Jaeger Client C-Sharp v1.0.2
 - Deployment: Jaeger deployed via Kubernetes but traced application is running on bare metal (Linux via WSL)

**What troubleshooting steps did you try?**
Code review.

**Additional context**
The following code update shows what I _think_ needs to be added to the `Jaeger.Senders.Grpc.GrpcSender` constructor to address the issue (notice we're passing the `channelOptions` list to the `Channel` constructor:
```csharp
        /// <param name="target">If empty it will use <see cref="DefaultCollectorGrpcTarget"/>.</param>
        /// <param name="credentials">If empty it will use <see cref="ChannelCredentials.Insecure"/>.</param>
        /// <param name="maxPacketSize">If 0 it will use <see cref="MaxPacketSize"/>.</param>
        public GrpcSender(string target, ChannelCredentials credentials, int maxPacketSize)
        {
            if (string.IsNullOrEmpty(target))
            {
                target = DefaultCollectorGrpcTarget;
            }

            if (credentials == null)
            {
                credentials = ChannelCredentials.Insecure;
            }

[...]

PR #226, opened 2021-08-16
Appropriately Set gRPC Client Max Send and Receive Message Length

js8080:
#225 

Even though the GrpcSender allowed specifying a maxPacketSize
parameter if that value was over 4MB, Spans over that 4MB limit
would not be recorded because the core gRPC client library has a
default max message length of 4MB. Now, the GrpcSender will set
the appropriate ChannelOptions to set the MaxSendMessageLength and
MaxReceiveMessageLength based on the given maxPacketSize.

Signed-off-by: Justin Stauffer <justin@epic.com>

### DIFF
--- a/src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
+++ b/src/Senders/Jaeger.Senders.Grpc/GrpcSender.cs
@@ -61,7 +61,13 @@ namespace Jaeger.Senders.Grpc
                 maxPacketSize = MaxPacketSize;
             }
 
-            _channel = new Channel(target, credentials);
+            var channelOptions = new List<ChannelOption>()
+            {
+                new ChannelOption(ChannelOptions.MaxSendMessageLength, maxPacketSize),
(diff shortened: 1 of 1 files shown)


Falco20019:
Thanks for your report and the PR. I will give it a look tomorrow (UTC+1 here). Can you please apply the DCO fix like described here: https://github.com/jaegertracing/jaeger-client-csharp/runs/3344037372

js8080:
> Thanks for your report and the PR. Can you please apply the DCO fix like described here: https://github.com/jaegertracing/jaeger-client-csharp/runs/3344037372
> 
> Just make sure that the mail used in the sign-off has to match the one used as author of the commit set in your got config. You can use a local got config and set the mail accordingly if you want to use your gmail instead of epic mail.
> 
> Besides that, the PR looks good to me and I will merge it once fixed.

I fixed the commit sign-off email. Also updated it in the PR description for consistency.

Falco20019 (approved):

PR #229, opened 2021-10-25
Feat update

EngRajabi:
- update all dependencies
- add netstandard2.1 and netstandard2.0
- update ApacheThrift to 0.15.0

- update all dependencies
- add netstandard2.1 and netstandard2.0
- update ApacheThrift to 0.15.0


Falco20019 (changes requested):
Thanks for your PR!

I'm fine with most changes. Doing the tests and samples on .NET 5 and updating those references should be OK. We just need to ensure to maximize the range of supported clients without enforcing updates on them.

Keep in mind, that the end client is always free to update those in the project.

Falco20019:
@EngRajabi Any more input? Are you fine with the proposed changes?

